PR #4, opened 2015-05-09
Pass parent window to file save dialog

lucaswerkmeister:
Requires rust-gnome/gtk#64.

Note: I haven’t been able to build this myself because Cargo doesn’t like me messing with its `.cargo/git/checkouts` directory. I *have* been able to compile the exact change in another program, though, so I’m fairly confident it should work.

### DIFF
--- a/src/text_viewer.rs
+++ b/src/text_viewer.rs
@@ -39,10 +39,10 @@ fn main() {
 
     window.add(&vbox);
 
-    open_button.connect_clicked(move |_| {
+    open_button.connect_clicked(|_| {
         // TODO move this to a impl?
         let file_chooser = gtk::FileChooserDialog::new(
-            "Open File", None, gtk::FileChooserAction::Open,
+            "Open File", Some(&window), gtk::FileChooserAction::Open,
             [("Open", gtk::ResponseType::Ok), ("Cancel", gtk::ResponseType::Cancel)]);
         if file_chooser.run() == gtk::ResponseType::Ok as i32 {
             let filename = file_chooser.get_filename().unwrap();


GuillaumeGomez:
Don't worry, I'll see your first PR (the code seems good so there shouldn't have any problem), then I'll restart travis build for this one.

lucaswerkmeister:
Great!

GuillaumeGomez:
Your first PR is merged, I restart travis on this one.

lucaswerkmeister:
Hm, looks like that didn’t work.

GuillaumeGomez:
Yep, moved value. I let you look what's wrong. ^^

lucaswerkmeister:
Perhaps because it’s a `move` closure? (Can’t test locally, git won’t pull in the merged PR yet.)

lucaswerkmeister:
Non-`move` closure seems to work…

GuillaumeGomez:
I'll wait @gkoz's opinion on that before doing anything more. But yes, it's normal that it works with non-move closure.

gkoz:
Any dialog improvements should probably build on [this PR](https://github.com/jeremyletang/rgtk/pull/246) which I let sit there because of some minor undecided issues (putting everything in one file may not be that nice, naming issues and a little nuance with the ButtonType).

I'm surprised you were able to capture a reference in a signal handler. This doesn't seem right. I'll look closer into this.

lucaswerkmeister:
Looks like your PR would actually make rust-gnome/gtk#64 redundant since your `parent`s are also `Option<&Window>`.

gkoz:
@lucaswerkmeister There's a bigger question there of what to do with the flags argument. Passing the parent as an `Option` removes the need to expose `GTK_DIALOG_MODAL` and `GTK_DIALOG_DESTROY_WITH_PARENT` flags. But there is another flag introduced in some GTK 3.x version which we may want to expose. This is a small mess that someone has to untangle creatively.

gkoz:
Yeah the fact that this PR builds is a bug in signals definitions. I let the lifetime elision put a wring lifetime in the `connect_*` functions time parameters - an implicit `&self` lifetime instead of `'static` which was intended. I think I'm going to have to fix all the signals now...

gkoz:
So https://github.com/rust-gnome/gtk/pull/65 will break this. You're going to have to clone the `window` and move the clone into the closure.

gkoz:
@lucaswerkmeister regarding the workflow: you don't need to mess with cargo checkouts, cargo allows you to [override dependencies](http://doc.crates.io/guide.html#overriding-dependencies) instead. There's a tool that can help you with that: https://github.com/gkoz/rust-gnome-manifest

lucaswerkmeister:
@gkoz Thanks! That’s good to know…

GuillaumeGomez:
Sorry I forgot this one. Is it good to merge or not yet ?

gkoz:
As it stands the example won't build.

GuillaumeGomez:
Ok, I'll wait for update then.

gkoz:
There've been some more dialog API changes bu it looks like this has been done.

PR #7, opened 2015-05-10
Restore gtktest

gkoz:
Let's see if Appveyor picks this up


GuillaumeGomez:
And you also put back all the old gtktest example code ! That's a wonderful day ! =D

PR #9, opened 2015-05-11
Use correct tab indexes in signal handler

mitaa:
Capturing the index by value does only work if tabs are added and removed in LIFO order.
Depends on https://github.com/rust-gnome/gtk/pull/73

I'm not quite sure about the `'static` bound on `Widget`, but I hope it's ok..

### DIFF
--- a/src/notebook.rs
+++ b/src/notebook.rs
@@ -17,8 +17,8 @@ impl NoteBook {
         }
     }
 
-    fn create_tab<'a, Widget: gtk::WidgetTrait>(&mut self, title: &'a str, widget: &Widget) ->
-            Option<u32> {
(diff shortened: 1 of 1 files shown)


gkoz:
Nice catch!

src/notebook.rs
```diff
@@ -38,7 +38,11 @@ impl NoteBook {
         };
 
         let notebook_clone = self.notebook.clone();
-        button.connect_clicked(move |_| notebook_clone.remove_page(index as i32));
+        let widget_clone = widget.clone();
```
gkoz: If you don't need `notebook` and `widget` after this, you can just move them into the closure without cloning.

GuillaumeGomez:
Thanks for your work ! I merge !

PR #12, opened 2015-05-16
Add a builder example to show the basics

tOkeshu:
I was playing with the `Builder` widget but could not make it work.
I decided to write a basic example so someone could help me fix what is going on and eventually add it to the example repo.

If you run `cargo build` you should see the following error:
```rust
src/builder_basics.rs:15:5: 15:22 error: the type of this value must be known in this context
src/builder_basics.rs:15     window.show_all();
                             ^~~~~~~~~~~~~~~~~
```

The glade file is taken from a [Python GTK tutorial](https://python-gtk-3-tutorial.readthedocs.org/en/latest/builder.html).

If we make this work, I should be able to add 1 or 2 examples showing how to use the builder.
Tell me what you think. :smile:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,3 +33,6 @@ name = "text_viewer"
 
 [[bin]]
 name = "treeview"
+
+[[bin]]
+name = "builder_basics"
--- a/src/builder_basics.rs
+++ b/src/builder_basics.rs
@@ -1,0 +1,40 @@
+//! # Builder Basics Sample
+//!
+//! This sample demonstrates how to use the builder with a simple glade file
+
+extern crate gtk;
+
+#[cfg(feature = "gtk_3_10")]
+mod example {
+    use gtk;
+    use gtk::traits::*;
+    use gtk::signal::Inhibit;
+    use gtk::widgets::Builder;
+    use gtk::Window;
+
+    pub fn main() {
+        gtk::init();
+        let builder = Builder::new_from_file("./builder_basics.glade").unwrap();
+        let window: Window = builder.get_object("window1").unwrap();
+
+        window.connect_delete_event(|_, _| {
+            gtk::main_quit();
+            Inhibit(true)
+        });
+
+        window.show_all();
+        gtk::main();
+    }
+}
+
+#[cfg(feature = "gtk_3_10")]
+fn main() {
+    example::main()
+}
+
+#[cfg(not(feature = "gtk_3_10"))]
+fn main() {
+    println!("This example only work with GTK 3.10 and later");
+    println!("Did you forget to build with `--features gtk_3_10`?");
+}
+
--- a/README.md
+++ b/README.md
@@ -6,6 +6,12 @@ A few rust-gnome examples. To build, just do:
 > cargo build
 ```
 
+or to enable GTK 3.10 examples as well:
+
+```Shell
+> cargo build --features gtk_3_10
+```
+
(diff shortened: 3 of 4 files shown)


Cargo.toml
```diff
@@ -7,9 +7,11 @@ authors = ["The Rust-gnome Project Developers"]
 libc = "0.1"
 glib = "^0"
 gdk = "^0"
-gtk = "^0"
 cairo-rs = "^0"
 
+[dependencies.gtk]
+features = ["gtk_3_10", "cairo_1_12"]
```
gkoz: We can't accept changes that prevent us from building the examples crate with older versions of GTK.
As this whole example seems to depend on 3.10 I suggest not touching the features in `Cargo.toml` and putting the whole logic into a submodule gated on `gtk_3_10` like this:
```Rust
extern crate glib;
extern crate gtk;

#[cfg(feature = "gtk_3_10")]
mod a {
    use gtk;
    // ...
    pub fn main() {
        // ...
    }
}

#[cfg(feature = "gtk_3_10")]
fn main() { a::main() }

#[cfg(not(feature = "gtk_3_10"))]
fn main() { }
```

GuillaumeGomez:
@tOkeshu: Hey, thanks for your PR ! However, we need examples wich can build on travis and appveyor in order to see if your work is building properly. So please follow @gkoz instructions to update your code. I'll merge your PR once this is done.

PS: Post a comment to notify me.

gkoz:
> error: the type of this value must be known in this context

The compiler doesn't know what we're building and needs out help like this:
```Rust
let window: gtk::Window = builder.get_object("window1").unwrap();
```

Currently there doesn't seem to be any runtime checking if the type is the correct one. This will probably be fixed in the upcoming object wrappers reform outlined in https://github.com/rust-gnome/gtk/pull/59

Cargo.toml
gkoz: Alternatively, we could add the [`add_from_file`](https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-add-from-file) method and avoid depending on 3.10 altogether.

tOkeshu: While I would prefer to have `add_from_file`, I think that would be a good thing to have an example using GTK 3.10, so next contributors can have something they can copy for other examples requiring 3.10.

tOkeshu:
I updated the patch to follow the module "scoped" to GTK 3.10.
I also modified a bit the example and updated the README.

Tell me if that's good for you @GuillaumeGomez and @gkoz :)

GuillaumeGomez:
@tOkeshu: Very good job ! It's perfectly fine for me. Is it good for you too @gkoz ?

src/builder_basics.rs
```diff
@@ -0,0 +1,41 @@
+//! # Builder Basics Sample
+//!
+//! This sample demonstrates how to use the builder with a simple glade file
+
+extern crate glib;
```
gkoz: We don't need this import, do we? Let's not confuse anyone who might cargo-cult this into their own projects :)

gkoz:
Aside of the last small nit, :+1: good job.

tOkeshu:
Fixed.

Thank you for your help @GuillaumeGomez and @gkoz.

GuillaumeGomez:
No problem ! Thank you for contributing ! Don't hesitate to help more if you have time ! =D

PR #21, opened 2015-07-02
Check the return value of gtk::init

gkoz:


GuillaumeGomez:
I guess it's a good thing to show "the good way". :p Thanks for adding it !

PR #24, opened 2015-07-18
Update dependencies

GuillaumeGomez:


GuillaumeGomez:
cc @gkoz

GuillaumeGomez:
cc @pnkfelix

gkoz:
Looks good. We'll probably want to set the default branch to `crate` [here](https://github.com/rust-gnome/examples/settings) so people would not assume that we want them to use git dependencies.

GuillaumeGomez:
It would be the opposite of all other repos. So as you prefer.

gkoz:
As there's no other distribution channel for binary crates, I don't see any way around it: new users should see the preferred version and there might be a message for contributors somewhere -- that you need to open PRs against `master`.

GuillaumeGomez:
I'll do it that way. I'll also update crates.

gkoz:
Keeping the crates dependency?

GuillaumeGomez:
I created a non-crate branch. Take a look! :-)

PR #28, opened 2015-07-31
Added an example of multithreaded program

koniahin:
Added an example of multithreaded program that uses thread local storage key.


GuillaumeGomez:
Thanks a lot!

cc @gkoz

gkoz:
:+1:

PR #43, opened 2015-10-21
Fix TreeIter breakage

gkoz:


gkoz:
This needs to be merged for https://github.com/gtk-rs/gtk/pull/176 to come out green.

src/treeview.rs
```diff
@@ -88,16 +85,14 @@ fn main() {
...
-        right_store.append(&mut iter, None);
+    for i in 0..10 {
+        let iter = right_store.append(None);
         right_store.set_value(&iter, 0, &value);
 
-        let mut child_iter = gtk::TreeIter::new();
-
-        right_store.append(&mut child_iter, Some(&iter));
-        right_store.set_string(&child_iter, 0, "I'm a child node");
+        for _ in (0..i) {
```
GuillaumeGomez:
```Rust
for _ in 0..i {
```

instead.

GuillaumeGomez:
Travis doesn't like this one. :/

src/treeview.rs
gkoz: Hm, right

gkoz:
The two PRs depend on each other so one of them has to be merged first anyway.

GuillaumeGomez:
Ok, then let's start with this one.

PR #45, opened 2015-11-03
Handle response idiomatically

tundrax:


GuillaumeGomez:
Hi, thanks for your PR !

However, I don't agree with your changes. We changed it this way because we didn't want to panic.

cc @gkoz

tundrax:
@GuillaumeGomez, thanks for the feedback. This is my first commit to the Rust project, thus I may lack credible argumentation. I thought if you are building a GTK app and can't `init` it, that has to be a certain `panic` point.

GuillaumeGomez:
No it doesn't, and that's why we didn't `panic` in `gtk-rs` either (we're still talking about this last point). When the initialization fails, the user can do whatever he wants, so, in here, he does return from the `main` function and the error message is less ugly.

Internally, I'm against panicking to let userd handle it however they want. If gtk isn't available, why the program should crash ? They might want to use another lib if this one fail (we never know ^^).

gkoz:
@tundrax I feel you might've been slightly misled by the book. While convenient to use in code samples and prototypes, panicing is not considered idiomatic. The use of `unwrap` and `expect` as means of error handling is frowned upon. I guess, we're emphasizing here that although failure to initialize GTK is fatal it is not completely unexpected and shouldn't "crash" the app.

tundrax:
@gkoz @GuillaumeGomez thanks for clarifying response. Please ignore the PR.

PR #51, opened 2015-11-27
Crate

gkoz:


gkoz:
This doesn't make sense. `gtk` build fails before `gtk-sys` is ever built. What

GuillaumeGomez:
I like dark magic too. :D

PR #54, opened 2015-12-28
Update to last rustc version

GuillaumeGomez:
cc @gkoz


src/cairo_threads.rs
```diff
@@ -163,8 +164,8 @@ fn draw_initial(format: Format, width: i32, height: i32) -> (Box<[u8]>, i32) {
     (buf, image.get_stride())
 }
 
-fn draw_slow(cr: &Context, delay: u32, x: f64, y: f64, radius: f64) {
-    thread::sleep_ms(delay);
+fn draw_slow(cr: &Context, delay: u64, x: f64, y: f64, radius: f64) {
```
gkoz: I guess `delay` should be `Duration` then?

gkoz:
I think we push to `pending` first and then could cherry-pick into the other branches if you'd like this to not wait for new crates.

GuillaumeGomez:
As you prefer, I don't have preference. I just want to avoid disturbing your current work.

src/cairo_threads.rs
GuillaumeGomez: I find it easier to understand this way. It's like a little wrapping.

gkoz: If the standard library "suggests" we take advantage of richer types i.e. adopt the more specific `Duration` type it seems logical to do this fully.

gkoz:
This shouldn't matter and the commit could be cherry-picked in any direction. But generally I'd prefer consistency ;)

gkoz:
I think it would be simpler to reopen this against the `pending` branch.

PR #58, opened 2016-01-13
Signals API adjustment

gkoz:
Objects are passed to handlers by reference now.
The first parameter type is the same as the type, `connect_*` was called on.


GuillaumeGomez:
I find it better this way. Thanks @gkoz.

PR #62, opened 2016-01-18
Tree view image

GuillaumeGomez:
cc @gkoz 

Currently, I have issues to put the image in the first column. Can you take a look at it when you have time please?


GuillaumeGomez:
The commit history is really weird...

gkoz:
If you want to put pixbufs in the column, you have to tell that to the tree store:

```patch
diff --git a/src/treeview.rs b/src/treeview.rs
index c5a1b67..5c45eb0 100644
--- a/src/treeview.rs
+++ b/src/treeview.rs
@@ -7,6 +7,7 @@ extern crate gtk;
 extern crate gdk;
 
 use gtk::prelude::*;
+use glib::types::StaticType;
 
 fn append_text_column(tree: &gtk::TreeView) {
     let column = gtk::TreeViewColumn::new();
@@ -93,7 +94,7 @@ fn main() {
 
     // right pane
     /*let right_tree = gtk::TreeView::new();
-    let right_column_types = [glib::Type::BaseBoxed, glib::Type::String];
+    let right_column_types = [gdk::pixbuf::Pixbuf::static_type(), glib::Type::String];
     let right_store = gtk::TreeStore::new(&right_column_types);
     let renderer = gtk::CellRendererPixbuf::new();
     let col = gtk::TreeViewColumn::new();
```

But it looks like this still can't work until we have proper `Value` bindings.

PR #66, opened 2016-01-23
Add hint

GuillaumeGomez:


gkoz:
Now that `FileError` is exported properly this can be 
```rust
            if e.matches::<glib::FileError>() == Some(glib::FileError::Noent) {
                println!("Try relaunching this example from the same level as \
                         the `resources` directory.");
            }
```

I wonder though if `matches` should be changed to `typed` for consistency with `Value`...

GuillaumeGomez:
And we should also think to make errors' name more verbose. Quite difficult to understand for now.

gkoz:
Yeah, the `FileError` is a mess. It should look closer to `std::io::ErrorKind`. Maybe there could even be conversions between them...

GuillaumeGomez:
http://blog.burntsushi.net/rust-error-handling/#error-handling-with-a-custom-type ?

gkoz:
A single enum works for a closed hierarchy. We have an open one here: any library can define additional error domains.

gkoz:
So, the above code would now be
```rust
            if e.typed::<glib::FileError>() == Some(glib::FileError::Noent) {
                println!("Try relaunching this example from the same level as \
                         the `resources` directory.");
            }
```

GuillaumeGomez:
Updated.

gkoz:
Thanks

PR #70, opened 2016-02-06
treeview: get_selected returns a model, rename the variable to reflect this

silvio:
Signed-off-by: Silvio Fricke <silvio.fricke@gmail.com>


GuillaumeGomez:
Hi, thanks for your PR (the signed-off is useless by the way)! I don't find it better, but since I'm the one who wrote it, my opinion is certainly biaised.

cc @gkoz

gkoz:
I guess it's difficult to argue against making the name match the type.

silvio:
thanks!

PR #71, opened 2016-02-13
Use new version features

gkoz:


Cargo.toml
```diff
@@ -19,7 +16,7 @@ git = "https://github.com/gtk-rs/gdk"
 git = "https://github.com/gtk-rs/gtk"
 
 [features]
-gtk_3_10 = []
+gtk_3_10 = ["gtk/3.10"]
 gtk_3_14 = ["gtk_3_10"]
```
EPashkin: This don't need "gtk/3.14" for completeness sake? Or to be removed as unused?

gkoz: Ideally `gtk_3_14` should not be there, I put it in when branching appveyor into 3.6 and 3.14 to try avoiding some confusion. Now it's all confusing anyway so might as well remove 3_14 but it has to start with CI scripts changes.

gkoz:
Looks like the version features merge fallout is fixed now.

GuillaumeGomez:
Thanks!

PR #74, opened 2016-03-04
Depend on gdk-pixbuf and use Duration

gkoz:
Closes #54


EPashkin:
Seems linked to wrong Issue

gkoz:
AFAIR that PR used to replace deprecated `sleep_ms` in a slightly different way but it's hard to tell now and it had been opened against the wrong branch anyway.

PR #75, opened 2016-03-08
Evolve treeview example

gkoz:


src/treeview.rs
```diff
@@ -44,21 +44,8 @@ fn main() {
...
-    });
-
     for _ in 0..10 {
-        let iter = left_store.append();
-        left_store.set_value(&iter, 0, &"I'm in a list".to_value());
-
-        // select this row as a test
-        //
-        left_selection.select_path(&left_store.get_path(&iter));
+        left_store.insert_with_values(None, &[0], &[&"I'm in a list"]);
```
GuillaumeGomez: I think that a comment which explains what happens here would be very appreciated.

PR #76, opened 2016-03-11
Add more simple treeview example

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/src/simple_treeview.rs
+++ b/src/simple_treeview.rs
@@ -1,24 +1,46 @@
 //! # TreeView Sample
 //!
-//! This sample demonstrates how to create a TreeView with either a ListStore or TreeStore.
+//! This sample demonstrates how to create a TreeView with a ListStore.
 
-extern crate glib;
 extern crate gtk;
 
 use gtk::prelude::*;
 use gtk::{
-    CellRendererText, TreeStore, TreeView, TreeViewColumn,
+    CellRendererText, Label, ListStore, Orientation, TreeView, TreeViewColumn,
     Window, WindowPosition, WindowType
 };
+
+fn create_and_fill_model() -> ListStore {
+    // Creation of a model with two rows.
+    let model = ListStore::new(&[u32::static_type(), String::static_type()]);
+
+    // Filling up the tree view.
+    let entries = &["Michel", "Sara", "Liam", "Zelda", "Neo", "Octopus master"];
+    for (i, entry) in entries.iter().enumerate() {
+        model.insert_with_values(None, &[0, 1], &[&(i as u32 + 1), &entry]);
+    }
+    model
+}
 
 fn append_column(tree: &TreeView, id: i32) {
     let column = TreeViewColumn::new();
     let cell = CellRendererText::new();
 
     column.pack_start(&cell, true);
-    // we create the column at the position `id`
+    // Association of the view's column with the model's `id` column.
     column.add_attribute(&cell, "text", id);
     tree.append_column(&column);
+}
+
+fn create_and_setup_view() -> TreeView {
+    // Creating the tree view.
+    let tree = TreeView::new();
+
+    tree.set_headers_visible(false);
+    // Creating the two columns inside the view.
+    append_column(&tree, 0);
+    append_column(&tree, 1);
+    tree
 }
 
 fn main() {
@@ -37,40 +59,39 @@ fn main() {
         Inhibit(false)
     });
 
-    // we create a view
-    let tree = TreeView::new();
-    // we create a model with two rows
-    let store = TreeStore::new(&[String::static_type(), i32::static_type()]);
+    // Creating a vertical layout to place both tree view and label in the window.
+    let vertical_layout = gtk::Box::new(Orientation::Vertical, 0);
 
-    // when a line is selected, this signal is called
+    // Creation of the label.
+    let label = Label::new(None);
+
+    let tree = create_and_setup_view();
+
+    let model = create_and_fill_model();
+    // Setting the model into the view.
+    tree.set_model(Some(&model));
+    // Adding the view to the layout.
+    vertical_layout.add(&tree);
+    // Same goes for the label.
+    vertical_layout.add(&label);
+
+    // The closure responds to selection changes by connection to "::cursor-changed" signal,
+    // that gets emitted when the cursor moves (focus changes).
     tree.connect_cursor_changed(move |tree_view| {
         let selection = tree_view.get_selection();
         if let Some((model, iter)) = selection.get_selected() {
-            // We are now getting back the values from the line corresponding to the
+            // Now getting back the values from the row corresponding to the
             // iterator `iter`.
             //
             // The `get_value` method do the conversion between the gtk type and Rust.
-            println!("Hello '{}' from line {}",
-                     model.get_value(&iter, 0).get::<String>().unwrap(),
-                     model.get_value(&iter, 1).get::<i32>().unwrap() + 1);
+            label.set_text(&format!("Hello '{}' from row {}",
+                                    model.get_value(&iter, 1).get::<String>().unwrap(),
+                                    model.get_value(&iter, 0).get::<u32>().unwrap()));
         }
     });
 
-    // we set the model into the view
-    tree.set_model(Some(&store));
-    tree.set_headers_visible(false);
-    // we create the two columns inside the view
-    append_column(&tree, 0);
-    append_column(&tree, 1);
-
-    // we fill the treeview
-    let entries = &["Michel", "Sara", "Liam", "Zelda", "Neo", "Octopus master"];
-    for (i, entry) in entries.iter().enumerate() {
-        store.insert_with_values(None, None, &[0, 1], &[&entry, &(i as i32)]);
-    }
-
-    // we add the view to the window
-    window.add(&tree);
+    // Adding the layout to the window.
+    window.add(&vertical_layout);
 
     window.show_all();
     gtk::main();
--- a/src/simple_treeview.rs
+++ b/src/simple_treeview.rs
@@ -1,0 +1,77 @@
+//! # TreeView Sample
+//!
+//! This sample demonstrates how to create a TreeView with either a ListStore or TreeStore.
+
+extern crate glib;
+extern crate gtk;
+
+use gtk::prelude::*;
+use gtk::{
+    CellRendererText, TreeStore, TreeView, TreeViewColumn,
+    Window, WindowPosition, WindowType
+};
+
+fn append_column(tree: &TreeView, id: i32) {
+    let column = TreeViewColumn::new();
+    let cell = CellRendererText::new();
+
+    column.pack_start(&cell, true);
+    // we create the column at the position `id`
+    column.add_attribute(&cell, "text", id);
+    tree.append_column(&column);
+}
+
+fn main() {
+    if gtk::init().is_err() {
+        println!("Failed to initialize GTK.");
+        return;
+    }
+
+    let window = Window::new(WindowType::Toplevel);
+
+    window.set_title("Simple TreeView example");
+    window.set_position(WindowPosition::Center);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    // we create a view
+    let tree = TreeView::new();
+    // we create a model with two rows
+    let store = TreeStore::new(&[String::static_type(), i32::static_type()]);
+
+    // when a line is selected, this signal is called
+    tree.connect_cursor_changed(move |tree_view| {
+        let selection = tree_view.get_selection();
+        if let Some((model, iter)) = selection.get_selected() {
+            // We are now getting back the values from the line corresponding to the
+            // iterator `iter`.
+            //
+            // The `get_value` method do the conversion between the gtk type and Rust.
+            println!("Hello '{}' from line {}",
+                     model.get_value(&iter, 0).get::<String>().unwrap(),
+                     model.get_value(&iter, 1).get::<i32>().unwrap() + 1);
+        }
+    });
+
+    // we set the model into the view
+    tree.set_model(Some(&store));
+    tree.set_headers_visible(false);
+    // we create the two columns inside the view
+    append_column(&tree, 0);
+    append_column(&tree, 1);
+
+    // we fill the treeview
+    let entries = &["Michel", "Sara", "Liam", "Zelda", "Neo", "Octopus master"];
+    for (i, entry) in entries.iter().enumerate() {
+        store.insert_with_values(None, None, &[0, 1], &[&entry, &(i as i32)]);
+    }
+
+    // we add the view to the window
+    window.add(&tree);
+
+    window.show_all();
+    gtk::main();
+}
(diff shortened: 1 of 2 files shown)


src/simple_treeview.rs
```diff
@@ -0,0 +1,72 @@
...
+    let store = TreeStore::new(&[String::static_type(), i32::static_type()]);
+
+    // when a line is selected, this signal is called
+    tree.connect_cursor_changed(move |tree_view| {
+        let selection = tree_view.get_selection();
+        if let Some((model, iter)) = selection.get_selected() {
+            println!("Hello '{}' from line {}",
+                     model.get_value(&iter, 0).get::<String>().unwrap(),
+                     model.get_value(&iter, 1).get::<i32>().unwrap() + 1);
+        }
```
gkoz: The thing I dislike about some of our examples is that they showcase features randomly.
The app should be coherent in accomplishing some goal. Since this function has no purpose the reader lacks any context that would help understand what happens here and why.

GuillaumeGomez: I wanted to link from the tutorial to our examples. But that's well saw, I'll add comments!

GuillaumeGomez:
I added some comments.

gkoz:
My objection to purposelessness still stands. Currently `connect_cursor_changed` doesn't solve any problem, even a silly imaginary one. The reader's brain is disengaged, they're coasting past disconnected feature demonstrations. `println!` is a symptom of this because printing to console can't serve any purpose in a graphical app.

GuillaumeGomez:
I'll link with a label then.

GuillaumeGomez:
Updated.

src/simple_treeview.rs
```diff
@@ -0,0 +1,93 @@
...
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    // we create a vertical layout to place both tree view and label in the window
+    let vertical_layout = gtk::Box::new(gtk::Orientation::Vertical, 0);
+
+    // we create the label, but because of the closure, we need to wrap it inside
+    // a Rc<RefCell<>> to be able to use it outside the closure
+    let label = Rc::new(RefCell::new(Label::new(Some(""))));
```
gkoz: Has https://github.com/GuillaumeGomez/process-viewer/commit/2878a17979aac74d5e70a81d4644393aab0b38d1 been for nothing?

Also if it takes option why not None then?

GuillaumeGomez: Btw, how comes the `new` takes an `Option` whereas the `set_text` doesn't?

GuillaumeGomez:
Indeed, sorry, totally forgot the `clone` method...

src/simple_treeview.rs
gkoz: There must be reasons I suppose but at least this doesn't appear to be our fault, this is consistent with documentation and `Gtk-3.0.gir`. :stuck_out_tongue:

gkoz:
Take another look at the first example [here](https://en.wikibooks.org/wiki/GTK%2B_By_Example/Tree_View/Columns_and_Renderers). It separates `create_and_fill_model` and `create_view_and_model` because the point of MVC is disentangling them. The model init, treeview init and signal connection should not be intermixed randomly.

 `append_column` being a screen away is a distraction. We're not after maximal DRY here.

Some terminology should be kept consistent with the upstream (here and in the tutorial):
- `ListStore` is a *model*
- signals get *emitted*, *signal handlers* get *connected* to signals, the handler or closure is what gets *called*

src/simple_treeview.rs
```diff
@@ -0,0 +1,90 @@
...
+    Window, WindowPosition, WindowType
+};
+
+fn append_column(tree: &TreeView, id: i32) {
+    let column = TreeViewColumn::new();
+    let cell = CellRendererText::new();
+
+    column.pack_start(&cell, true);
+    // we create the column at the position `id`
+    column.add_attribute(&cell, "text", id);
```
gkoz: position where?

src/simple_treeview.rs
```diff
@@ -0,0 +1,90 @@
...
+    CellRendererText, Label, TreeStore, TreeView, TreeViewColumn,
+    Window, WindowPosition, WindowType
+};
+
+fn append_column(tree: &TreeView, id: i32) {
+    let column = TreeViewColumn::new();
+    let cell = CellRendererText::new();
+
+    column.pack_start(&cell, true);
+    // we create the column at the position `id`
```
gkoz: position where?

src/simple_treeview.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    // we create the label, but because of the closure, we need to wrap it inside
+    // a Rc<Cell<>> to be able to use it outside the closure
+    let label = Label::new(None);
+    let label_copy = label.clone();
+
+    // we create the tree view
+    let tree = TreeView::new();
+    // we create a model with two rows
+    let store = TreeStore::new(&[String::static_type(), i32::static_type()]);
```
gkoz: Is there a reason the numbers column is the second one, has a signed type and starts with 0?

gkoz: Or that this is not a `ListStore`?

src/simple_treeview.rs
GuillaumeGomez: How would you improve this sentence? I don't see how to explain differently we're creating a new column at position `id`.

gkoz: We're not. We're associating the column with the `id` column of the model.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+extern crate glib;
+extern crate gtk;
+
+use gtk::prelude::*;
+use gtk::{
+    CellRendererText, Label, ListStore, TreeView, TreeViewColumn,
+    Window, WindowPosition, WindowType
+};
+
+fn append_column(tree: &TreeView, id: i32) {
```
gkoz: This function can be closer to `create_and_setup_view`

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+    let column = TreeViewColumn::new();
+    let cell = CellRendererText::new();
+
+    column.pack_start(&cell, true);
+    // We associate the view's column with the model's `id` column.
```
gkoz: "We" everywhere is distracting.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+    column.add_attribute(&cell, "text", id);
+    tree.append_column(&column);
+}
+
+fn create_and_fill_model() -> ListStore {
+    // We create a model with two rows.
+    let store = ListStore::new(&[String::static_type(), u32::static_type()]);
```
gkoz: Still puzzled by the order. I've never seen a list with the "№" column on the right-hand side.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+    window.set_title("Simple TreeView example");
+    window.set_position(WindowPosition::Center);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    // We create a vertical layout to place both tree view and label in the window.
+    let vertical_layout = gtk::Box::new(gtk::Orientation::Vertical, 0);
```
gkoz: Import `Orientation`?

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    // We create the label, but because of the closure, we need to wrap it inside
+    // a Rc<Cell<>> to be able to use it outside the closure.
```
gkoz: Stale comment.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+    let label = Label::new(None);
+    let label_copy = label.clone();
```
gkoz: I'd rather not call it a copy because there's no copying involved.

src/simple_treeview.rs
GuillaumeGomez: You're hard on business!

src/simple_treeview.rs
GuillaumeGomez: The order doesn't matter but if you prefer, I can put it on the left.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+            //
+            // The `get_value` method do the conversion between the gtk type and Rust.
+            label_copy.clone()
+                      .set_text(&format!("Hello '{}' from line {}",
+                                         model.get_value(&iter, 0).get::<String>().unwrap(),
+                                         model.get_value(&iter, 1).get::<u32>().unwrap() + 1));
+        }
+    });
+
+    let store = create_and_fill_model();
```
gkoz: No point calling it `store` here either.

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+    let tree = create_and_setup_view();
+    // When a line is selected, this signal is emitted so we connect to it through this method.
+    tree.connect_cursor_changed(move |tree_view| {
+        let selection = tree_view.get_selection();
+        if let Some((model, iter)) = selection.get_selected() {
+            // We are now getting back the values from the line corresponding to the
+            // iterator `iter`.
+            //
+            // The `get_value` method do the conversion between the gtk type and Rust.
+            label_copy.clone()
```
gkoz: Why `clone`?

src/simple_treeview.rs
GuillaumeGomez: It's the "easy" way to say "internally, we're using g_ref to increment the number of references pointing to this object to avoid it going to the void when the destructor is called". :)

src/simple_treeview.rs
```diff
@@ -0,0 +1,101 @@
...
+                      .set_text(&format!("Hello '{}' from line {}",
+                                         model.get_value(&iter, 0).get::<String>().unwrap(),
+                                         model.get_value(&iter, 1).get::<u32>().unwrap() + 1));
```
gkoz: Why make numbers in the list and the label inconsistent?

src/simple_treeview.rs
gkoz: I suppose it is an occasion to stress the point that each `Label` or whatever is a smart pointer, cloning it creates another reference to the same object and our library provides automatic resource management. Those things certainly aren't obvious.

src/simple_treeview.rs
gkoz: I believe it will look more natural with numbers on the left.

gkoz:
I'm pretty sure now that since we don't tweak any of the rederer's properties one `CellRendererText` is sufficient.

src/simple_treeview.rs
```diff
@@ -0,0 +1,98 @@
+//! # TreeView Sample
+//!
+//! This sample demonstrates how to create a TreeView with either a ListStore.
```
gkoz: Redundant "either"

src/simple_treeview.rs
```diff
@@ -0,0 +1,98 @@
...
+    let cell = CellRendererText::new();
+
+    column.pack_start(&cell, true);
+    // Association of the view's column with the model's `id` column.
+    column.add_attribute(&cell, "text", id);
+    tree.append_column(&column);
+}
+
+fn create_and_fill_model() -> ListStore {
+    // Creation of a model with two rows.
```
gkoz: It was "We create" now it's "Creation of" why not "Create"?

GuillaumeGomez: I like using an unpersonal talk way.

gkoz:
> I'm pretty sure now that since we don't tweak any of the rederer's properties one CellRendererText is sufficient.

But maybe it would give the wrong idea for other cases, I don't know...

src/simple_treeview.rs
gkoz: Any embellishments of the most straightforward form "Create a model..." make the reader spend more cycles on this line than necessary. If that's the intention, fine.

src/simple_treeview.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    let model = create_and_fill_model();
+    // Setting the model into the view.
+    tree.set_model(Some(&model));
+    // Adding the view to the layout.
+    vertical_layout.add(&tree);
+    // Same goes for the label.
+    vertical_layout.add(&label);
+
+    // When a line is selected, this signal is emitted so we connect to it through this method.
```
gkoz: "line is selected." seems a bit vague, the closure seems to be called almost every time I click somewhere and the docs say "The position of the cursor (focused cell) has changed."

"this signal" Does the reader know about the `connect_$SIGNAL` convention? 

"so we" and "through this method" don't seem to convey any information.

src/simple_treeview.rs
```diff
@@ -0,0 +1,98 @@
...
+extern crate glib;
+extern crate gtk;
+
+use gtk::prelude::*;
+use gtk::{
+    CellRendererText, Label, ListStore, Orientation, TreeView, TreeViewColumn,
+    Window, WindowPosition, WindowType
+};
+
+fn append_column(tree: &TreeView, id: i32) {
```
gkoz: This function could be closer to `create_and setup_view`

gkoz:
Oh, and it's "row" not "line".

src/simple_treeview.rs
GuillaumeGomez: > "line is selected." seems a bit vague, the closure seems to be called almost every time I click somewhere and the docs say "The position of the cursor (focused cell) has changed."

That's the point of this signal. But if you prefer, I can write "clicked" instead.

GuillaumeGomez: > "this signal" Does the reader know about the connect_$SIGNAL convention?

I'd prefer assume the reader has at least the basics in gtk when reading this code.

gkoz: You can change focus with keyboard though. Maybe "when cursor moves (focus changes)".

GuillaumeGomez: That's even worst. I think the best might just be "when the selected cell changes".

gkoz: > I'd prefer assume the reader has at least the basics in gtk when reading this code.

gtk doesn't have `connect_cursor_changed`.

gkoz: How is repeating the documentation literally -- worst and paraphrasing it subjectively better?

GuillaumeGomez: It's not. But it says what it does, and that's what we're trying to do after all.

gkoz: I'd put it like this:
Respond to selection changes by connecting to the "::cursor-changed" signal, that gets emitted when the cursor moves (focus changes).

GuillaumeGomez: Perfect, I take it!

src/simple_treeview.rs
GuillaumeGomez: You mean the name or the content?

gkoz: Placement. They're separated by the model function unnecessarily.

src/simple_treeview.rs
```diff
@@ -0,0 +1,99 @@
+//! # TreeView Sample
+//!
+//! This sample demonstrates how to create a TreeView with a ListStore.
+
+extern crate glib;
```
gkoz: Oh I missed this one. It's not needed is it?

GuillaumeGomez: Well saw!

PR #82, opened 2016-03-28
Revert "Fix call TreeModel.get_path"

gkoz:
This kind of is a public-facing branch so let's do this by the book not by force-pushing.


EPashkin:
Agree

EPashkin:
@gkoz I can merge, so merge it.

PR #85, opened 2016-04-22
Fix building with upper gtk versions

EPashkin:
IMHO `gtk_3_18` feature better changed to `gtk_3_16` for consistency
Also `defaults` commented as example

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,10 @@ git = "https://github.com/gtk-rs/gdk"
 git = "https://github.com/gtk-rs/gtk"
 
 [features]
+#default = ["gtk_3_18"]
 gtk_3_10 = ["gtk/v3_10"]
-gtk_3_18 = ["gtk_3_10"]
+gtk_3_16 = ["gtk_3_10", "gtk/v3_16"]
+gtk_3_18 = ["gtk_3_16"] #for CI tools
 
 [[bin]]
 name = "basic"


GuillaumeGomez:
Agreed. Thanks @EPashkin.

cc @gkoz

EPashkin:
Not sure how to change to v3_16: just lower feature or lower installed version in travis.

GuillaumeGomez:
Just the feature should be enough.

EPashkin:
Lowered feature

EPashkin:
Any 3_18 still won't builds due `Gesture`

EPashkin:
Repush to restart builds

GuillaumeGomez:
@EPashkin: If you want to restart builds, just ask to me or to @gkoz.

gkoz:
Just removing `gtk_3_18` will break appveyor for all other repos.

EPashkin:
Add it again as `gtk_3_18 = ["gtk_3_16"]` ?

EPashkin:
We can also add `gtk_latest = ["gtk_3_16"]` and update all other repos to it.

gkoz:
`latest` isn't too good when looked at from the point of view of the CI build matrix.
I think I'd prefer `gtk_3_18 = ["gtk_3_16"]` but with a comment pointing out that it's used by CI tools. But then there remains little reason to have `gtk_3_16`. You just can't win it seems.

EPashkin:
IMHO Main reason for "gtk_3_16" is example of right way of adding versions.

Cargo.toml
```diff
@@ -19,8 +19,10 @@ git = "https://github.com/gtk-rs/gdk"
 git = "https://github.com/gtk-rs/gtk"
 
 [features]
+#default = ["gtk_3_18"]
```
gkoz: What are we saying by including this line? It could be interpreted in unpredictable ways.

EPashkin: Its example for setting upper version by default

gkoz:
Anyway, bikeshedding this is too boring. Thanks!

PR #86, opened 2016-04-23
Spin button signals

EPashkin:
WIP.
Example using https://developer.gnome.org/gtk3/stable/GtkSpinButton.html#GtkSpinButton-input after https://github.com/gtk-rs/gtk/pull/317
Seems returning  `GTK_INPUT_ERROR` still applied `new_value` (don't changing it equals `new_value = 0.;`), then not sure that we need it.

cc @gkoz 

Edit: `GTK_INPUT_ERROR` working normal and it depends of `update_policy`.
This old code for future reference
```rust
            Ok(value) if value >= 99. => {
                println!("circular");
                *new_value = 10.;
                Err(()) //Not working
            }
```

### DIFF
--- a/src/gtktest.rs
+++ b/src/gtktest.rs
@@ -4,6 +4,7 @@ extern crate gtk;
 extern crate gdk;
 
 use gtk::prelude::*;
+use gtk::SpinButtonUpdatePolicy;
 use gdk::enums::modifier_type;
 
 /// Expands to its argument if GTK+ 3.10 support is configured and to `()` otherwise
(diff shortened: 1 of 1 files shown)


EPashkin:
Returning `FALSE` from trampoline not applied `new_value`

EPashkin:
Seems update on error depends on `update_policy` which `GTK_UPDATE_ALWAYS` by default.

EPashkin:
Seems closure return value need changing to something like `Result<f64, bool>`.
Where `Err(true)` mean `GTK_INPUT_ERROR`, `Err(false)` -> FALSE.
But I don't like how it looks.

EPashkin:
It won't work with `GTK_UPDATE_ALWAYS` as we needed return `f64` too :cry:

EPashkin:
From `gtk` code `gtk_entry_get_text` don't need `Option`. @gkoz, what you think?

EPashkin:
Updated to reflect https://github.com/gtk-rs/gtk/pull/318

EPashkin:
@GuillaumeGomez, please, restart builds

GuillaumeGomez:
@EPashkin: Restarted.

src/gtktest.rs
```diff
@@ -150,6 +151,24 @@ fn main() {
...
+        match text.parse::<f64>() {
+            Ok(value) if value >= 90. => {
+                println!("circular right");
+                Some(Ok(10.))
+            }
+            Ok(value) if value <= 10. => {
+                println!("circular left");
+                Some(Ok(90.))
+            }
+            Ok(_) => None,
```
gkoz: If the value is parsed already why not return it?

EPashkin: Was to check that changing ignored with FALSE.
Change to return value?

gkoz: I don't think returning `None` is going to behave differently from `Some(Ok(value))` -- the library will parse the text and produce the same value. So just returning it feels more straightforward.

EPashkin:
Updated

EPashkin:
Can I update build script to show stats?

GuillaumeGomez:
@EPashkin: I think it could be a good idea actually.

EPashkin:
Asked in wrong PR :(
Added stats to gtk's `make gir`

gkoz:
Thanks!

PR #88, opened 2016-04-24
Add menu bar example

GuillaumeGomez:
cc @gkoz


EPashkin:
No check items and others?

GuillaumeGomez:
I wanted to make this example as light as possible but maybe it's too light now haha.

src/menu_bar.rs
```diff
@@ -0,0 +1,85 @@
...
+    h_box.pack_start(&menu_bar, false, false, 0);
+    h_box.pack_start(&label, true, true, 0);
+    window.add(&h_box);
+    window.show_all();
+
+    about.connect_activate(move |_| {
+        let p = Dialog::new_with_buttons(Some("About"),
+                                         Some(&window),
+                                         gtk::DIALOG_MODAL,
+                                         &[("Ok", gtk::ResponseType::Ok as i32)]);
```
gkoz: Huh. I guess we've got a problem: a breaking change, which doesn't actually break the build. This new enum wasn't supposed to be convertible via `as i32` but it still is. Might have to make the `__Nonexhaustive` member non-empty to prevent this.
The proper conversion would be `ResponseType::Ok.into()`.

EPashkin: Dialog still needed `i32` not `gtk::ResponseType`?

gkoz: It supports arbitrary response codes and until `ResponseType` accommodates that it needs `i32`. Additionally, gir doesn't have a facility to substitute `ResponseType` for `i32` where we want it to :(

gkoz: To be clear, right now `ResponseType::Ok as i32 == 4` while for the type formerly known as `ResponseType` it's `gtk_sys::GtkResponseType::Ok as i32 == -5` and that is what `.into()` will return.

EPashkin: May be it be error after https://github.com/rust-lang/rust/issues/3868 ?

EPashkin: Also make `__Nonexhaustive(())` will break `as i32`

gkoz: Yes, we'll probably make `__Nonexhaustive(())`.

EPashkin: Then I make 3 PR's: 
examples: change `as i32` to `.into()` for all generated enums,
gir: change generation,
gtk: regen

gkoz: :+1:

src/menu_bar.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let label = Label::new(Some("MenuBar example"));
+
+    h_box.pack_start(&menu_bar, false, false, 0);
+    h_box.pack_start(&label, true, true, 0);
+    window.add(&h_box);
+    window.show_all();
+
+    about.connect_activate(move |_| {
+        let p = Dialog::new_with_buttons(Some("About"),
```
gkoz: Aren't you supposed to use `AboutDialog` here?

GuillaumeGomez: I am.

src/menu_bar.rs
EPashkin: Gtk changes in https://github.com/EPashkin/rust-gnome-gtk/tree/block_enum_to_i32

GuillaumeGomez:
@EPashkin: Happier? ;)

EPashkin:
@GuillaumeGomez rebase please

EPashkin:
Hm, now github as ` New changes since you last viewed`

EPashkin:
Looks great.
May be better hide "License" button in "About" as it empty?

EPashkin:
Also closing not work in "About", seems it result of
```
Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.
```

EPashkin:
In `gtktest` all normal.

src/menu_bar.rs
```diff
@@ -0,0 +1,105 @@
...
+    window.add(&h_box);
+    window.show_all();
+
+    about.connect_activate(move |_| {
+        let p = AboutDialog::new();
+        p.set_authors(&["gtk-rs developers"]);
+        p.set_website_label(Some("gtk-rs"));
+        p.set_website(Some("http://gtk-rs.org"));
+        p.set_comments(Some("What a menu bar example!"));
+        p.set_copyright(Some("All rights reserved to us (of course!)"));
```
gkoz: Come on, the license is MIT

gkoz:
> In gtktest all normal with About.

GTK asks for a transient window and it's set in gtktest.

src/menu_bar.rs
GuillaumeGomez: Okay, I put it...

GuillaumeGomez:
I changed copyright (less fun now...) and added parent window.

EPashkin:
About still has not working "License" and "Close" button, in "gtktest" license is hidden and close working.
And rebase please

src/menu_bar.rs
```diff
@@ -0,0 +1,106 @@
...
+    folder_box.pack_start(&folder_image, false, false, 0);
+    folder_box.pack_start(&folder_label, true, true, 0);
+    folder_entry.add(&folder_box);
+    menu.append(&file_entry);
+    menu.append(&folder_entry);
+    menu.append(&check_entry);
+    menu.append(&about);
+    menu.append(&quit);
+    file.set_submenu(Some(&menu));
+    menu_bar.append(&file);
```
gkoz: We should try to use Glade (UI metadata editor). No one is expected to build UIs programmatically these days.

GuillaumeGomez: Anyone can use Glade. I think it's a good thing to show how it works programatically. However, we could provide all examples in Glade version.

gkoz: There's a chance we're unaware of some obstacles to using Glade, especially around menus and actions. The GTK learning materials push strongly toward using Glade, doing a lot of these manipulations manually seems unidiomatic. So we need to prioritize making sure the Glade way works.

GuillaumeGomez:
So, do we merge this PR?

gkoz:
I'd very much like it if the menu (or the whole UI) would be defined in XML.

GuillaumeGomez:
And I said that I'd do it, but in a separated example.

GuillaumeGomez:
Actually, we should do this for all examples.

gkoz:
Frankly, teaching users to do this stuff manually is a disservice. This is not what they need to learn to be efficient at making GTK apps.

GuillaumeGomez:
All the other examples have been written by hand. This one isn't an exception. The point of these examples is to show how do it, either they want to do it by using GLADE or not.

gkoz:
I'm afraid this hasn't been how to do it for years now. We just sort of weren't paying attention. It's unfortunate that the existing examples don't match the modern ways but it's not an excuse to not learn new things and keep rehashing the old stuff.

GuillaumeGomez:
And so that's why I proposed to have both ways.

EPashkin:
@GuillaumeGomez seems no changes to fixing my comment (not tested).
> "About" still has not working "License" and "Close" button, in "gtktest" license is hidden and close working.
> And rebase please

EPashkin:
@gkoz direct way generating allow to show how to use library, so I prefer have only one Glade example and other same that this.

GuillaumeGomez:
> @gkoz direct way generating allow to show how to use library, so I prefer have only one Glade example and other same that this.

You're more extreme than me! ;) Providing an example in both GLADE and full "hand-made" source code seems a good compromise.

gkoz:
Okay, let's have the manual example. You still need to make loading of the image work regardless of `$CWD`.

EPashkin:
I changed my options on using glade: goal of examples is show how do some task and better that example concentrated to this task, so if you need do some complex preparation then better use any way to simplify it.
But my option about this example still unaffected as example for working with menu.

@gkoz Can we include image to code as resource at compile time and use it?

GuillaumeGomez:
> @gkoz Can we include image to code as resource at compile time and use it?

It's possible, I did it in C++. However I have no clue how to do so in Rust...

gkoz:
> goal of examples is show how do some task and better that example concentrated to this task

Normally programmatic UI initialization isn't something one should want to accomplish and examples suggesting that approach distract from learning the idiomatic way.

>  Can we include image to code as resource at compile time and use it?

Presumably using `glib-compile-resources` and then doing something [like this](https://github.com/gtk-rs/glib/issues/120#issuecomment-214694739) should work. We might consider making a convenience function to achieve that with less boilerplate.

GuillaumeGomez:
Updated. Do you still have the issues @EPashkin?

EPashkin:
Thanks, @GuillaumeGomez.
All good except picture, but it can be done later, as it better changed to many examples.

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
...
+//!
+//! This sample demonstrates how to create a TreeView with either a ListStore or TreeStore.
+
+extern crate gtk;
+
+use gtk::{
+    AboutDialog, BoxExt, CheckMenuItem, ContainerExt, DialogExt, IconSize, Image, Inhibit, Label,
+    Menu, MenuBar, MenuItem, MenuItemExt, MenuShellExt, WidgetExt, WidgetSignals,
+    Window, WindowExt, WindowPosition, WindowType
+};
```
gkoz: We don't import traits.

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
...
+    window.set_title("MenuBar example");
+    window.set_position(WindowPosition::Center);
+    window.set_size_request(400, 400);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    let h_box = gtk::Box::new(gtk::Orientation::Vertical, 10);
```
gkoz: `v_box`?

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
...
+        p.set_website_label(Some("gtk-rs"));
+        p.set_website(Some("http://gtk-rs.org"));
+        p.set_authors(&["Gtk-rs developers"]);
+        p.set_title("About!");
+        p.set_transient_for(Some(&window));
+        p.run();
+        p.destroy();
+    });
+    check_entry.connect_toggled(|w| {
+        w.set_label(if w.get_label().unwrap() == "Unchecked" {
```
gkoz: Is it possible to actually test whether the checkbox is checked?

GuillaumeGomez: It is. Do you prefer it this way?

src/menu_bar.rs
GuillaumeGomez: Well saw!

src/menu_bar.rs
GuillaumeGomez: I'll replace by `gtk::prelude::*` after what we experienced recently.

src/menu_bar.rs
gkoz: Yes it will make the currently present bug impossible.

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
...
+    });
+
+    let h_box = gtk::Box::new(gtk::Orientation::Vertical, 10);
+
+    let menu = Menu::new();
+    let menu_bar = MenuBar::new();
+    let file = MenuItem::new_with_label("File");
+    let about = MenuItem::new_with_label("About");
+    let quit = MenuItem::new_with_label("Quit");
+    let file_entry = MenuItem::new();
```
gkoz: Why are items called 'entries'?

src/menu_bar.rs
GuillaumeGomez: What bug?

src/menu_bar.rs
GuillaumeGomez: It's a menu entry. So... Well, my explanation seems good enough. :D

gkoz: Can we avoid misleading names please? It says it's an 'item' right there.

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
...
+        p.set_website_label(Some("gtk-rs"));
+        p.set_website(Some("http://gtk-rs.org"));
+        p.set_authors(&["Gtk-rs developers"]);
+        p.set_title("About!");
+        p.set_transient_for(Some(&window));
+        p.run();
+        p.destroy();
+    });
+    check_entry.connect_toggled(|w| {
+        w.set_label(if w.get_active() == true {
```
gkoz: `== true` is redundant

GuillaumeGomez: I wondered about it for a while between being explicit or redundant.

GuillaumeGomez:
Updated.

src/menu_bar.rs
```diff
@@ -0,0 +1,107 @@
+//! # TreeView Sample
+//!
+//! This sample demonstrates how to create a TreeView with either a ListStore or TreeStore.
```
gkoz: Oh and these seem out of place.

GuillaumeGomez: Urg, silly me!

GuillaumeGomez:
Updated.

gkoz:
Thanks!

PR #91, opened 2016-04-28
Adjust to ImageSurface::get_data change

gkoz:
Fix breakage after https://github.com/gtk-rs/cairo/pull/75


EPashkin:
:+1: checked that it compiles with `cairo#75` and `cairo_threads` is working

Issue #92, opened 2016-05-04
Glade


GuillaumeGomez:
As discussed with @gkoz, I'll rewrite all examples using Glade and we'll move current examples to a test repository.

GuillaumeGomez:
Done since a while now so I close it.

PR #93, opened 2016-05-07
Glade

GuillaumeGomez:
Translate some examples to Glade.

Once this merged, the original examples (the non-Glade code) will be moved into its own repository (to keep tests on "normal" code).

Part of #92

### DIFF
--- a/src/gtktest.glade
+++ b/src/gtktest.glade
@@ -174,12 +174,12 @@ Leonard McCoy</property>
                     </child>
                     <child>
                       <object class="GtkLinkButton" id="link_button">
-                        <property name="label" translatable="yes">button</property>
+                        <property name="label" translatable="yes">rust-lang.org</property>
                         <property name="visible">True</property>
                         <property name="can_focus">True</property>
                         <property name="receives_default">True</property>
                         <property name="relief">none</property>
-                        <property name="uri">www.rust-lang.org</property>
+                        <property name="uri">https://www.rust-lang.org/</property>
                       </object>
                       <packing>
                         <property name="expand">False</property>
--- a/src/gtktest.glade
+++ b/src/gtktest.glade
@@ -1,0 +1,552 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.16.1 -->
+<interface>
+  <requires lib="gtk+" version="3.10"/>
+  <object class="GtkAdjustment" id="scale_adjustment">
+    <property name="upper">100</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAdjustment" id="spin_adjustment">
+    <property name="upper">100</property>
+    <property name="step_increment">1</property>
+    <property name="page_increment">10</property>
+  </object>
+  <object class="GtkAboutDialog" id="dialog">
+    <property name="can_focus">False</property>
+    <property name="type_hint">dialog</property>
+    <property name="program_name">Glade</property>
+    <property name="website">http://gtk-rs.org</property>
+    <property name="authors">James T. Kirk
+Spock
+Leonard McCoy</property>
+    <property name="artists">James T. Kirk</property>
+    <child internal-child="vbox">
+      <object class="GtkBox" id="aboutdialog-vbox1">
+        <property name="can_focus">False</property>
+        <property name="orientation">vertical</property>
+        <property name="spacing">2</property>
+        <child internal-child="action_area">
+          <object class="GtkButtonBox" id="aboutdialog-action_area1">
+            <property name="can_focus">False</property>
+            <property name="layout_style">end</property>
+          </object>
+          <packing>
+            <property name="expand">False</property>
+            <property name="fill">True</property>
+            <property name="pack_type">end</property>
+            <property name="position">0</property>
+          </packing>
+        </child>
+        <child>
+          <placeholder/>
+        </child>
+      </object>
+    </child>
+  </object>
+  <object class="GtkImage" id="image2">
+    <property name="visible">True</property>
+    <property name="can_focus">False</property>
+    <property name="icon_name">edit-clear</property>
+  </object>
+  <object class="GtkWindow" id="window">
+    <property name="can_focus">False</property>
+    <property name="title" translatable="yes">Yeah a beautiful window with gtk !</property>
+    <property name="window_position">center</property>
+    <child>
+      <object class="GtkFrame" id="frame">
+        <property name="visible">True</property>
+        <property name="can_focus">False</property>
+        <property name="border_width">10</property>
+        <property name="label_xalign">0</property>
+        <property name="shadow_type">none</property>
+        <child>
+          <object class="GtkAlignment" id="alignment1">
+            <property name="visible">True</property>
+            <property name="can_focus">False</property>
+            <property name="left_padding">12</property>
+            <child>
+              <object class="GtkBox" id="_box">
+                <property name="visible">True</property>
+                <property name="can_focus">False</property>
+                <property name="border_width">5</property>
+                <property name="orientation">vertical</property>
+                <property name="spacing">10</property>
+                <child>
+                  <object class="GtkInfoBar" id="info_bar">
+                    <property name="visible">True</property>
+                    <property name="app_paintable">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="show_close_button">True</property>
+                    <child internal-child="action_area">
+                      <object class="GtkButtonBox" id="infobar-action_area1">
+                        <property name="can_focus">False</property>
+                        <property name="spacing">6</property>
+                        <property name="layout_style">end</property>
+                        <child>
+                          <placeholder/>
+                        </child>
+                        <child>
+                          <placeholder/>
+                        </child>
+                        <child>
+                          <placeholder/>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">False</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child internal-child="content_area">
+                      <object class="GtkBox" id="infobar-content_area1">
+                        <property name="can_focus">False</property>
+                        <property name="spacing">16</property>
+                        <child>
+                          <placeholder/>
+                        </child>
+                        <child>
+                          <placeholder/>
+                        </child>
+                        <child>
+                          <placeholder/>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">False</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <placeholder/>
+                    </child>
+                  </object>
+                  <packing>
+                    <property name="expand">False</property>
+                    <property name="fill">True</property>
+                    <property name="position">0</property>
+                  </packing>
+                </child>
+                <child>
+                  <object class="GtkBox" id="v_box">
+                    <property name="visible">True</property>
+                    <property name="can_focus">False</property>
+                    <property name="spacing">10</property>
+                    <child>
+                      <object class="GtkSwitch" id="switch">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">0</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkSwitch" id="switch2">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="active">True</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">1</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkCheckButton" id="check_button">
+                        <property name="label" translatable="yes">checkbutton</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">False</property>
+                        <property name="xalign">0</property>
+                        <property name="draw_indicator">True</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">2</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkLinkButton" id="link_button">
+                        <property name="label" translatable="yes">button</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">True</property>
+                        <property name="relief">none</property>
+                        <property name="uri">www.rust-lang.org</property>
+                      </object>
+                      <packing>
+                        <property name="expand">False</property>
+                        <property name="fill">True</property>
+                        <property name="position">3</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkSpinButton" id="spin_button">
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="adjustment">spin_adjustment</property>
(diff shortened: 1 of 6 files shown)


src/gtktest.rs
```diff
@@ -3,295 +3,213 @@
...
-        $bl
+mod example {
+    use gtk;
+    use gtk::prelude::*;
+    use gtk::Builder;
+    use gdk::enums::modifier_type;
+
+    /// Expands to its argument if GTK+ 3.10 support is configured and to `()` otherwise
+    #[cfg(not(feature = "gtk_3_10"))]
+    macro_rules! with_gtk_3_10 {
```
gkoz: Isn't this macro redundant now?

src/gtktest.rs
```diff
@@ -3,295 +3,213 @@
...
+        if gtk::init().is_err() {
+            println!("Failed to initialize GTK.");
+            return;
+        }
+        println!("Major: {}, Minor: {}", gtk::get_major_version(), gtk::get_minor_version());
+        let glade_src = include_str!("gtktest.glade");
+        let builder = Builder::new_from_string(glade_src);
+
+        let window: gtk::Window = builder.get_object("window").unwrap();
+        let button: gtk::Button = builder.get_object("button").unwrap();
```
gkoz: Better import `Window`, `Button` and so on. Also the code would probably be easier to understand if calls to `connect` were closer to `get_object`. This `button` is used like 30 lines from where it's created.

src/gtktest.rs
```diff
@@ -3,295 +3,213 @@
...
+                Ok(value) if value <= 10. => {
+                    println!("circular left");
+                    Some(Ok(90.))
+                }
+                Ok(value) => Some(Ok(value)),
+                Err(_) => Some(Err(())),
+            }
+        });
+
+        let entry_clone = entry.clone();
```
gkoz: Looks like the `clone!` macro is needed here too.

src/sync_widgets.rs
```diff
@@ -4,46 +4,53 @@
 
 extern crate gtk;
 
-use gtk::prelude::*;
-
-fn main() {
-    if gtk::init().is_err() {
-        println!("Failed to initialize GTK.");
-        return;
+#[cfg(feature = "gtk_3_10")]
```
gkoz: If the only reason for this is `Builder::new_from_string` we should rather bind the `add_from_string` method, which does not require 3.10.

src/sync_widgets.rs
```diff
@@ -4,46 +4,53 @@
...
+        if gtk::init().is_err() {
+            println!("Failed to initialize GTK.");
+            return;
+        }
+        let glade_src = include_str!("sync_widgets.glade");
+        let builder = Builder::new_from_string(glade_src);
+
+        let window: gtk::Window = builder.get_object("window").unwrap();
+        let spin_button: gtk::SpinButton = builder.get_object("spin_button").unwrap();
+        spin_button.set_range(0.0, 130.0);
```
gkoz: This should be possible to set from Glade.

GuillaumeGomez: Certainly, but I can't find where. :-/

src/gtktest.rs
GuillaumeGomez: Worst: it was unused. Strange there wasn't a warning about it.

GuillaumeGomez:
@gkoz: I think I addressed all your comments.

src/sync_widgets.glade
```diff
@@ -0,0 +1,43 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.16.1 -->
+<interface>
+  <requires lib="gtk+" version="3.10"/>
```
gkoz: This requirement doesn't apply now?

gkoz:
The sync_widgets examples also doesn't need gtk 3.10 I assume.

Can you set up adjustments in Glade instead of in the code?

src/gtktest.rs
```diff
@@ -3,295 +3,189 @@
...
-        let dialog = gtk::AppChooserDialog::new_for_content_type(Some(&window), gtk::DIALOG_MODAL,
-            "sh");
+    pub fn sub_main() {
+        if gtk::init().is_err() {
+            println!("Failed to initialize GTK.");
+            return;
+        }
+        println!("Major: {}, Minor: {}", gtk::get_major_version(), gtk::get_minor_version());
+        let glade_src = include_str!("gtktest.glade");
+        let builder = Builder::new_from_string(glade_src);
```
EPashkin: As @gkoz says this not work in 3.10, while `add_from_string` works.
IMHO if no other controls need 3.10, we can
```
let builder = if cfg!(feature = "gtk_3_10")  {
 Builder::new_from_string(glade_src)
} else {
let builder = Builder::new();
builder.add_from_string(glade_src)
}
```
and remove other switches

src/gtktest.rs
```diff
@@ -3,295 +3,189 @@
...
+            //entry.set_text("Clicked!");
+            let dialog = AppChooserDialog::new_for_content_type(Some(&window), gtk::DIALOG_MODAL,
+                "sh");
+
+            dialog.run();
+            dialog.destroy();
+        }));
+
+        let button_about: Button = builder.get_object("button_about").unwrap();
+        // use a plain function instead of a closure
```
EPashkin: Its TODO?

src/gtktest.rs
```diff
@@ -3,295 +3,189 @@
...
+        let entry: Entry = builder.get_object("entry").unwrap();
+        let spinner: Spinner = builder.get_object("spinner").unwrap();
+        spinner.start();
+
+        /*info_bar.connect(signals::Response::new(|response_id| {
+            info_bar.hide()
+        }));*/ //TODO: Why does this not work?
+
+        let scale: Scale = builder.get_object("scale").unwrap();
+        scale.set_range(0., 100.);
```
EPashkin: To glade?

src/sync_widgets.glade
GuillaumeGomez: Arf!

src/gtktest.rs
GuillaumeGomez: In this example, a few other things need 3.10 gtk version, so just changing this wouldn't be enough.

src/gtktest.rs
GuillaumeGomez: No idea, it was like that so I kept it.

src/gtktest.rs
GuillaumeGomez: Like I already said, I didn't find how to do it in glade directly.

src/gtktest.rs
gkoz: `cfg!` can't prevent build errors. That's the reason the `with_gtk_3_10` macro exists. Anyway it's not like the add methods are deprecated.

src/gtktest.rs
gkoz: This example sort of shows that you can also use functions as signal handlers.

src/gtktest.rs
gkoz: It's not a big deal, none of us know GTK very well. It's puzzling that you'd go on not knowing and offload the googling to someone else.

src/gtktest.rs
EPashkin: But it don't use function as signal handler, compare it with `drawing_area.connect_draw(draw_fn);` from `cairotest.rs`

src/gtktest.rs
```diff
@@ -3,295 +3,189 @@
...
+        println!("Major: {}, Minor: {}", gtk::get_major_version(), gtk::get_minor_version());
+        let glade_src = include_str!("gtktest.glade");
+        let builder = Builder::new_from_string(glade_src);
+
+        let window: Window = builder.get_object("window").unwrap();
+        let entry: Entry = builder.get_object("entry").unwrap();
+        let spinner: Spinner = builder.get_object("spinner").unwrap();
+        spinner.start();
+
+        /*info_bar.connect(signals::Response::new(|response_id| {
```
EPashkin: Remove this comment?
In not worked as `window.show_all();` called after connect,
same effect can be done by glade if really needed. (not sure about it)

src/gtktest.rs
EPashkin: May be this help, change with `GtkAdjustment`?
http://stackoverflow.com/questions/30407433/gtkscale-draws-correctly-in-glade-not-in-the-final-app
Not checked.

GuillaumeGomez: > It's not a big deal, none of us know GTK very well. It's puzzling that you'd go on not knowing and offload the googling to someone else.

That's quite offending that you might I just let things like this in hope someone will do it for me. I let it like this because I wasn't able to find the way to do it.

src/gtktest.rs
EPashkin: > In this example, a few other things need 3.10 gtk version, so just changing this wouldn't be enough.

Strange, `gtktest` seems work after using `builder.add_from_string`.
Bad that `if cfg!` not work

EPashkin: `text_viewer` also works, despite `<requires lib="gtk+" version="3.10"/>`
May be it checked only in runtime with real gtk version.

EPashkin: .. and we can't just change building and assume that everything is fine

src/gtktest.rs
gkoz: You're right, then there's no point having the function. We don't really need those `println`s, they come from the time when I wasn't confident in array translations.

EPashkin: My problem is that comment saying "I'm use function instead closure"  but still use closure, so better remove comment.
I don't see anything wrong with debug prints in example.

src/gtktest.rs
gkoz: That's exactly what you're doing by claiming to have addressed all comments. Someone has to spend effort to find out if that is true and then solve the issue because clearly you're done with it.

GuillaumeGomez: I guess you're right. Sorry, I finally found out.

src/gtktest.rs
GuillaumeGomez: So what do you want me to do?

EPashkin: Sorry, was wrong, all normal here.

GuillaumeGomez:
Ok, so I updated. I removed `set_range` calls from code to add them to glade directly (damn adjustment thing). I removed unneeded comments. Do you see anything else?

src/gtktest.glade
```diff
@@ -0,0 +1,552 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!-- Generated with glade 3.16.1 -->
+<interface>
+  <requires lib="gtk+" version="3.10"/>
+  <object class="GtkAdjustment" id="adjustment1">
```
EPashkin: Not better 'scale_adjustment' ?

src/gtktest.rs
```diff
@@ -3,295 +3,182 @@
...
+        if gtk::init().is_err() {
+            println!("Failed to initialize GTK.");
+            return;
+        }
+        println!("Major: {}, Minor: {}", gtk::get_major_version(), gtk::get_minor_version());
+        let glade_src = include_str!("gtktest.glade");
+        let builder = Builder::new_from_string(glade_src);
+
+        let window: Window = builder.get_object("window").unwrap();
+        let entry: Entry = builder.get_object("entry").unwrap();
```
EPashkin: Can you move it down to use place?

src/gtktest.rs
```diff
@@ -3,295 +3,182 @@
...
+            //entry.set_text("Clicked!");
+            let dialog = AppChooserDialog::new_for_content_type(Some(&window), gtk::DIALOG_MODAL,
+                "sh");
+
+            dialog.run();
+            dialog.destroy();
+        }));
+
+        let button_about: Button = builder.get_object("button_about").unwrap();
+        button_about.connect_clicked(move |x| {
```
EPashkin: First click was good, second show many asserts, empty window and crash on mouse over.

src/gtktest.glade
EPashkin: I talked about id, other ids for GtkAdjustment also not good.

GuillaumeGomez: Since we don't use them, I didn't give them a "correct" name. I change it.

src/gtktest.rs
GuillaumeGomez: It comes from the fact that Glade creates only one instance of it, so we need to show/hide instead of destroy. I added a comment in the source code to explain it.

GuillaumeGomez:
Updated.

src/gtktest.rs
```diff
@@ -3,295 +3,184 @@
...
+    pub fn sub_main() {
+        if gtk::init().is_err() {
+            println!("Failed to initialize GTK.");
+            return;
+        }
+        println!("Major: {}, Minor: {}", gtk::get_major_version(), gtk::get_minor_version());
+        let glade_src = include_str!("gtktest.glade");
+        let builder = Builder::new_from_string(glade_src);
+
+        let window: Window = builder.get_object("window").unwrap();
```
EPashkin: Down to `let button` ?

src/sync_widgets.rs
```diff
@@ -4,37 +4,28 @@
...
-    let slider = gtk::Scale::new_with_range(gtk::Orientation::Horizontal, 0.0, 130.0, 1.0);
-
-    let hbox = gtk::Box::new(gtk::Orientation::Horizontal, 5);
-
-    hbox.set_homogeneous(true);
-    hbox.add(&spin_button);
-    hbox.add(&slider);
-
-    window.add(&hbox);
     window.show_all();
```
EPashkin: Not sure if it wrong but it always near `gtk::main();` after all connects

GuillaumeGomez: It's just to make life easier by avoid useless `clone` calls.

EPashkin: I just moved it down and has no problem. `window` used only in `window.connect_delete_event` so also `get_object` can move there.

GuillaumeGomez: Indeed, I moved it. Well saw!

EPashkin:
:+1: after fixing last commit name or split it to olds. Thanks.

GuillaumeGomez:
Forgot to squash :open_mouth:

EPashkin:
@GuillaumeGomez by way, you can use comment names like `fixup! Translate sync_widgets example to glade` and do rebase after squashing (if rebase.autoSquash = true set)

GuillaumeGomez:
Oh I didn't know. Thanks for the tip!

EPashkin:
* not `after squashing` but `after PR approved`

EPashkin:
IMHO this also helps to see what fixed

EPashkin:
Now last commit contains both `gtktest` and `sync_widgets` :(

GuillaumeGomez:
Ok, i rewrite the commits as well. :)

GuillaumeGomez:
@EPashkin: I rewrote commits. Better?

src/gtktest.glade
```diff
@@ -0,0 +1,552 @@
...
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkLinkButton" id="link_button">
+                        <property name="label" translatable="yes">button</property>
+                        <property name="visible">True</property>
+                        <property name="can_focus">True</property>
+                        <property name="receives_default">True</property>
+                        <property name="relief">none</property>
+                        <property name="uri">www.rust-lang.org</property>
```
EPashkin: https://www.rust-lang.org/

EPashkin: `www.rust-lang.org` was it label

GuillaumeGomez: You want me to change the label by the way?

EPashkin: Yes `button` not good.
IMHO label better without `https://` as tooltip shows full address.

GuillaumeGomez: Yes sir!

EPashkin:
Thanks

gkoz:
Looks good, except text_viewer also doesn't seem to need 3.10.

GuillaumeGomez:
@gkoz: Fixed.

gkoz:
Thanks for researching Glade! We can now think what can be done about ergonomics of this approach.

PR #96, opened 2016-06-14
Merge pending for gtk 0.1.0 release

gkoz:


GuillaumeGomez:
We're using a way too old rust version on appveyor. We should update it. I merge. Thanks @gkoz!

Issue #79, opened 2016-03-26
[Request] Gtk::Application example


vaurelios:
Would be nice have a example of usage `Gtk::Application` in rust, that shows what's recommended way to do this.

sdroege:
Here is a small example https://github.com/chergert/hello/blob/master/rust/src/main.rs

sdroege:
This is not really closed yet, an example in this repository would still be nice to have

sdroege:
Oh well, I filed the more general https://github.com/gtk-rs/examples/issues/140 for this now :)

PR #97, opened 2016-06-27
Add basic application example

tones111:
This is an attempt at addressing issue #79.  I'm new to rust and gtk so feedback on improvements are welcome.

The need for the new_app function is unfortunate, but all my attempts at using cfg! result in non-compiling code because the false branch of the conditional doesn't get optimized out early enough during compilation.

I probably missed something, but it's also unfortunate that the gio dependency is needed for the call to gtk::Application::new.  Is there any mechanism to access the constant through the gtk crate?

### DIFF
--- a/src/basic_application.rs
+++ b/src/basic_application.rs
@@ -1,0 +1,50 @@
+//! # Basic Application Sample
+//!
+//! This sample demonstrates how to create a GTK application with a toplevel `window`, set its title, size and position, how to add a `button` to this `window` and how to connect signals with actions.
+
+#![crate_type = "bin"]
+
+extern crate gio;
+extern crate gtk;
+
+use gtk::Builder;
+use gtk::prelude::*;
+
+const APP_ID: &'static str = "org.gtk-rs.basic_app";
+const APP_FLAGS: gio::ApplicationFlags = gio::APPLICATION_FLAGS_NONE;
+
+#[cfg(feature = "gtk_3_6")]
+fn new_app() -> Result<gtk::Application, ()> {
+    gtk::Application::new(Some(APP_ID), APP_FLAGS)
+}
+
+#[cfg(not(feature = "gtk_3_6"))]
+fn new_app() -> Result<gtk::Application, ()> {
+    gtk::Application::new(APP_ID, APP_FLAGS)
+}
+
+fn main() {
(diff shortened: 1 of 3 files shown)


src/basic_application.rs
```diff
@@ -0,0 +1,51 @@
...
+
+extern crate gio;
+extern crate gtk;
+
+use gtk::prelude::*;
+
+const APP_ID: &'static str = "org.gtk-rs.basic_app";
+const APP_FLAGS: gio::ApplicationFlags = gio::APPLICATION_FLAGS_NONE;
+
+#[cfg(feature = "gtk_3_10")]
```
EPashkin: In https://github.com/gtk-rs/gtk/blob/master/src/application.rs option `application_id` start from v3_6.
https://developer.gnome.org/gtk3/stable/GtkApplication.html#gtk-application-new says same

tones111:
I agree with you, but when building the examples the feature for the gtk version doesn't get passed to the compiler.  It sees the gtk_3_x features, but only gets passed a "gtk" feature.  Are you requesting I add a gtk_3_6 feature to the examples?

src/basic_application.rs
```diff
@@ -0,0 +1,51 @@
...
+            println!("clicked");
+        });
+        window.add(&button);
+
+        window.show_all();
+    });
+
+    // use -h to see usage options
+    let args: Vec<String> = std::env::args().collect();
+    let args: Vec<&str> = args.iter().map(|x| &x[..]).collect();
```
EPashkin: Why not do it in single pass?

EPashkin:
Yes.

EPashkin:
Bigger problem: examples PRs must go against `pending` branch, not master.
Recreate it pls.

src/basic_application.rs
GuillaumeGomez: Maybe because he wanted a `Vec<&str>` instead of a `Vec<String>`?

EPashkin: I mean `std::env::args().map(|x| &x[..]).collect();`

GuillaumeGomez: In your code, `x` doesn't live long enough.

EPashkin: Oh, that reason for multiple lines. Forgot about that.

PR #105, opened 2016-12-19
Rename grid child properties to "cell"

EPashkin:
Preparation for merging https://github.com/gtk-rs/gtk/pull/416.
Normal that it fails.


GuillaumeGomez:
Thanks!

PR #112, opened 2017-03-11
Use crates with git replacements in `pending` branch

EPashkin:
This seems allow use crated `Cargo.toml` in master branch (https://github.com/EPashkin/gtk/blob/master/Cargo.toml) and remove need in `crate` branch at all.

Works well with `.cargo/config/path` overrides. Ex. my local version with https://github.com/gtk-rs/gdk/pull/140 and https://github.com/gtk-rs/gtk/pull/462 rebased on my master produces only 2 equal warnings (dependency really added in gdk's PR)
```
warning: path override for crate `gdk` has altered the original list of
dependencies; the dependency on `gobject-sys` was either added or
modified to not match the previously resolved version
```
with one `warning: package replacement is not used: https://github.com/rust-lang/crates.io-index#gdk:0.5.2
` per replaced crate in `.cargo/config/path`

⚠️ WIP. Currently linked to my forks of gtk-rs as versions in masters don't upgraded

cc @GuillaumeGomez

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -55,3 +55,21 @@ name = "simple_treeview"
 
 [[bin]]
 name = "menu_bar"
+
+[replace]
+"gtk:0.1.2" = { git = 'https://github.com/EPashkin/gtk' }
+"gdk:0.5.2" = { git = 'https://github.com/EPashkin/gdk' }
+"gdk-pixbuf:0.1.2" = { git = 'https://github.com/EPashkin/gdk-pixbuf' }
+"pango:0.1.2" = { git = 'https://github.com/EPashkin/pango' }
+"cairo-rs:0.1.2" = { git = 'https://github.com/EPashkin/cairo' }
+"cairo-sys-rs:0.3.3" = { git = 'https://github.com/EPashkin/cairo' }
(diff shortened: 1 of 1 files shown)


EPashkin:
As possible bonus:
`replacement` can contain don't published version without generating error, so publishing new `gtk-rs` maybe can work without errors in order examples, sys, cairo, ..., gtk.

GuillaumeGomez:
Wo, I didn't know such a feature existed in Cargo! Generally I'm not much in favor for new Cargo feature but that might be worth it. If it impacts user, please write a README somewhere in order to explain how they can use it please.

Otherwise, very well done (and found)!

EPashkin:
I will do it at evening, also needed add PRs to all gtk-rs crates (it prepared in my forks)

EPashkin:
Starting to add PR's to gtk-rs crates, IMHO almost all of it fails due old version in master

GuillaumeGomez:
Tell me if a restart is needed for CIs.

EPashkin:
You merged this 😭 I add other PR with fixes

GuillaumeGomez:
Arf, sorry. :-/

PR #114, opened 2017-03-12
Revert Cargo.toml to version before `replace`

EPashkin:
Doing it manually as there I can redo only last of 2 PRs

Part of https://github.com/gtk-rs/gtk/issues/465


GuillaumeGomez:
Do I wait or should I merge them as soon as you open them?

PR #116, opened 2017-03-23
WIP: Add drag and drop example

Susurrus:
This requires the `drag_targets` branch of `gtk` and `gdk` to be used to work at all. It's also not quite correct, but I'm not certain why yet, and these examples are a good test case to make sure that the work in `gtk` and `gdk` is correct. One shows an example of dragging a widget within an app onto another widget. The other will display metadata of anything dropped onto it.

### DIFF
--- a/src/drag_and_drop.rs
+++ b/src/drag_and_drop.rs
@@ -1,0 +1,69 @@
+extern crate gdk;
+extern crate gtk;
+
+use gtk::prelude::*;
+
+fn main() {
+    if gtk::init().is_err() {
+        println!("Failed to initialize GTK.");
+        return;
+    }
+
+    // Configure button as drag source
+    let button = gtk::Button::new_with_label("Drag here");
+    let targets = vec![
+        gtk::TargetEntry::new("STRING", gtk::TargetFlags::empty(), 0),
+        gtk::TargetEntry::new("text/plain", gtk::TargetFlags::empty(), 0),
+    ];
+    button.drag_source_set(gdk::MODIFIER_MASK,
+                           &targets,
+                           gdk::ACTION_COPY);
+    button.connect_drag_data_get(|_, _, s, _, _| {
+        println!("drag-data-get");
+        let data = "I'm data!";
+        s.set_text(data, data.len() as i32);
+    });
+    button.connect_drag_begin(|_, c| {
+        println!("drag-begin");
+        for t in c.list_targets() {
+            println!("{}", t.name());
+        }
+    });
+    button.connect_drag_end(|_, _| {
+        println!("drag-end");
+    });
+
+    // Configure label as drag destination
+    let label = gtk::Label::new("Drop here");
+    label.drag_dest_set(gtk::DEST_DEFAULT_ALL,
+                        &targets,
+                        gdk::ACTION_COPY);
+    label.connect_drag_data_received(|w, d, _, _, s, _, t| {
+        println!("drag-data-received");
+        if s.get_length() > 0 && s.get_format() == 8 {
+            w.set_text(&s.get_text().unwrap());
+        }
+        d.drag_finish(false, false, t);
+    });
+    button.connect_drag_drop(|_, _, _, _, _| {
+        println!("drag-drop");
+        true
+    });
+
+    // Pack widgets into the window and display everything
+    let hbox = gtk::Box::new(gtk::Orientation::Horizontal, 0);
+    hbox.pack_start(&button, true, true, 0);
+    hbox.pack_start(&label, true, true, 0);
+
+    let window = gtk::Window::new(gtk::WindowType::Toplevel);
+    window.add(&hbox);
+    window.show_all();
+
+    // GTK & main window boilerplate
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    gtk::main();
+}
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,12 +3,20 @@ name = "gtk-rs-examples"
 version = "0.0.1"
 authors = ["The Gtk-rs Project Developers"]
 
-[dependencies]
-glib = "^0"
-gdk = "^0"
-gdk-pixbuf = "^0"
-gtk = "^0"
-cairo-rs = "^0"
+[dependencies.glib]
+git = "https://github.com/gtk-rs/glib"
+
+[dependencies.cairo-rs]
+git = "https://github.com/gtk-rs/cairo"
+
+[dependencies.gdk-pixbuf]
+git = "https://github.com/gtk-rs/gdk-pixbuf"
+
+[dependencies.gdk]
+git = "https://github.com/gtk-rs/gdk"
+
+[dependencies.gtk]
+git = "https://github.com/gtk-rs/gtk"
 
 [features]
 #default = ["gtk_3_20"]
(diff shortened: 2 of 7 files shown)


src/text_drop.rs
```diff
@@ -0,0 +1,52 @@
...
+    let accepted = vec![
+        gtk::TargetEntry::new("text/uri-list", gtk::TargetFlags::empty(), 0),
+//        gtk::TargetEntry::new("text/plain", gtk::TargetFlags::empty(), 1),
+//        gtk::TargetEntry::new("STRING", gtk::TargetFlags::empty(), 2),
+    ];
+    text_view.drag_dest_set(gtk::DEST_DEFAULT_ALL,
+                            &accepted,
+                            gdk::ACTION_COPY | gdk::ACTION_MOVE); // Maybe this should be ACTION_ASK?
+    text_view.connect_drag_data_received(|_, d, _, _, s, tt, t| {
+        println!("-{}", tt);
```
GuillaumeGomez: We're generally not big fans of printing things in stdout when using a graphic application. Use a small window or any equivalent which allows to remain in the application to see the info please.

src/drag_and_drop.rs
```diff
@@ -0,0 +1,75 @@
...
+        println!("Source Drag Data Delete");
+        println!("Delete the data!");
+    });
+
+    // Configure label as drag destination
+    label.drag_dest_set(gtk::DEST_DEFAULT_ALL,
+                        &[],
+                        gdk::ACTION_COPY); // Maybe this should be ACTION_ASK?
+    label.connect_drag_motion(|w, d, _, _, _| {
+        println!("motion");
```
GuillaumeGomez: Same.

src/drag_and_drop.rs
```diff
@@ -0,0 +1,75 @@
...
+
+    window.add(&vbox);
+    window.show_all();
+
+    // Configure button as drag source
+    button.drag_source_set(gdk::BUTTON1_MASK | gdk::BUTTON2_MASK,
+                           &[],
+                           gdk::ACTION_COPY | gdk::ACTION_MOVE);
+    button.connect_drag_data_get(|_, _, s, _, _| {
+        println!("Source Drag Data Get");
```
GuillaumeGomez: Same.

src/drag_and_drop.rs
```diff
@@ -0,0 +1,75 @@
...
+        let data = "I'm data!";
+        s.set_text(data, data.len() as i32);
+    });
+    button.connect_drag_data_delete(|w, d| {
+        println!("Source Drag Data Delete");
```
GuillaumeGomez: Same.

src/drag_and_drop.rs
```diff
@@ -0,0 +1,75 @@
...
+    // Configure label as drag destination
+    label.drag_dest_set(gtk::DEST_DEFAULT_ALL,
+                        &[],
+                        gdk::ACTION_COPY); // Maybe this should be ACTION_ASK?
+    label.connect_drag_motion(|w, d, _, _, _| {
+        println!("motion");
+        true
+    });
+    label.connect_drag_drop(|w, d, _, _, _| {
+        println!("drop");
```
GuillaumeGomez: Same.

src/drag_and_drop.rs
```diff
@@ -0,0 +1,75 @@
...
+        if !d.list_targets().is_empty() {
+            return true;
+        }
+        false
+    });
+    label.connect_drag_data_received(|w, d, _, _, s, _, _| {
+        println!("received");
```
GuillaumeGomez: Same.

Susurrus:
@GuillaumeGomez Indeed, this initial commit was primarily testing the library code I'm working on so you guys could see what's going on as I'm testing the changes to GTK and GDK.

Here's a smaller singular example. But when I run it I get a critical error about an Atom as soon as the GUI starts up. I'm not certain what exactly is going on here, but it looks like it's a failure in the Atom code I'm working on for GDK. @EPashkin any ideas?

```
(.:10245): Gdk-CRITICAL **: gdk_atom_intern: assertion 'atom_name != NULL' failed
```

EPashkin:
@Susurrus it mean that you pass NULL pointer in this function.
Can you give me line that give you this error?

Susurrus:
That's the entire error message I see on stdout, is there a way to get more detailed debugging information?

Susurrus:
With the most recent changes I made to gtk-rs/gtk#472, I no longer get that error and I see the highlighting on the drop target, but `connect_drag_data_received` is never called, so I'm investigating that now.

EPashkin:
You can add some `println("{}: {}", file!(), line!());` to detect lines that causes error

Susurrus:
The most recent problem now is that setting the selection data doesn't work, calling `s.set_text` returns failure. I think this is why the `connect_drag_data_received` signal isn't being called, but I'm not certain about that. The code looks like:
```
    button.connect_drag_data_get(|_, _, mut s, _, _| {
        let data = "I'm data!";
        println!("{}", s.set_text(data, data.len() as i32));
    });
```

It'd also be nice is `s.set_text()` didn't need the length argument, but that's just a sidenote.

GuillaumeGomez:
> It'd also be nice is `s.set_text()` didn't need the length argument, but that's just a sidenote.

That's actually an excellent point!

EPashkin:
What returning `targets_include_text` before `set_text` attempt?
[Help](https://developer.gnome.org/gtk3/stable/gtk3-Selections.html#gtk-selection-data-set-text) says that text converted to target.
So need check `gtk_selection_data_get_target` and `gtk_selection_data_get_targets` too.

Susurrus:
I'm not too familiar with how drag and drop works with GTK+ as their docs aren't very rich in this area, but I don't think that should be necessary because a) that isn't done in the example I'm porting over to Rust and b) these two are configured with identical text targets, so there shouldn't be any checking necessary.

Susurrus:
For reference, this is the drag and drop example from the gtkmm project that I'm working off of: https://developer.gnome.org/gtkmm-tutorial/stable/sec-dnd-example.html.en

Susurrus:
Looks like there might be an error with `gdk::DragContext::list_targets()` as it crashes when called in the latest iteration. The `drag-begin` and `drag-end` signals otherwise seem to be working properly though.

EPashkin:
@Susurrus Maybe `list_target` now fixed with last glib.

Susurrus:
@EPashkin `list_targets` is still crashing for me even with the latest glib.

Susurrus:
Besides `list_targets` not working, `SelectionData::set_text()` isn't working within `WidgetExt::connect_drag_data_get()`. This is all that's required for this example, though the `list_targets` issue is still concerning.

EPashkin:
@Susurrus, Main problem with both of your examples that it must based and pulled to `pending` branch.
Still looking why `list_targets` panics and why no drag-icon and result.

EPashkin:
- `list_targets` can fixed with
```
impl GlibPtrDefault for Atom {
    type GlibType = ffi::GdkAtom;
}
```
but after this need checked all function accepts or returns arrays of `Atom`.
Currently don't remember any checked.
- `set_text` don't work due `targets_include_text` returning false
this IMHO can be partially fixed with
```
let list = button.drag_source_get_target_list().unwrap();
list.add_text_targets(0);
```
but really won't work. It added to list 7 text targets
```
UTF8_STRING
COMPOUND_TEXT
TEXT
STRING
text/plain;charset=utf-8
text/plain;charset=CP1251
text/plain
```
- main problem: `drag_source_set` add 2 `TargetEntry` with empty string. I still don't understand why.

EPashkin:
Also if pass `&[]` to `drag_source_set` and `add_text_targets` after it
closure of `connect_drag_data_get` can change text (it checked by `get_text`) while `targets_include_text` still return false

Susurrus:
I'll try to replicate what you've posted, as of right now I've pushed a rebased version so this is easier to work with. Not certain if you want me to open a new PR for merging this into the `pending` branch or not, so please let me know about that.

EPashkin:
@Susurrus, Yes you need add PRs to `pending` branch` and close old.

PR #118, opened 2017-03-28
Add simple clipboard example

Susurrus:
Ported from https://developer.gnome.org/gtkmm-tutorial/stable/sec-clipboard-examples.html.en

Current implementation works, but is blocking on gtk-rs/gtk#478 and gtk-rs/gdk#148 which are soon to be merged.

There is additional functionality that is blocking on gtk-rs/gtk#479, which hasn't been started, but it's commented out.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,6 +70,9 @@ name = "cairotest"
 name = "child-properties"
 
 [[bin]]
+name = "clipboard_simple"
+
+[[bin]]
 name = "clock"
 
 [[bin]]
--- a/src/bin/clipboard_simple.rs
+++ b/src/bin/clipboard_simple.rs
@@ -1,0 +1,163 @@
+//! Simple clipboard example
+//!
+//! From https://developer.gnome.org/gtkmm-tutorial/stable/sec-clipboard-examples.html.en
+extern crate gdk;
+extern crate gio;
+extern crate gtk;
+
+use std::cell::RefCell;
+use std::env::args;
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+// make moving clones into closures more convenient
+macro_rules! clone {
+    (@param _) => ( _ );
+    (@param $x:ident) => ( $x );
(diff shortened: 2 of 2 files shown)


GuillaumeGomez:
Code seems great. I'll just wait for the feature to be merged before this example if don't mind. Thanks anyway!

EPashkin:
Just as reminder:
https://github.com/gtk-rs/examples/blob/master/CONTRIBUTING.md

Susurrus:
Alright, rebased from `pending` and cleaned up everything. GTK has managed to add all the functionality that was missing for this example over the last 2 years, so now we can finally merge this puppy!

GuillaumeGomez:

Susurrus:
Did you mean to close this issue? I just updated this a few days ago so it's all ready to merge now.

GuillaumeGomez:
Sorry, I just looked at the opening date of the PR, not the last update...

GuillaumeGomez:
I think you need to update your code. :)

Susurrus:
Yeah, so looks like there were breaking changes on the GTK side. For example `Label::new()` changed from `new<'a, P: Into<Option<&'a str>>>(str: P) -> Label` to `new(str: Option<&str>) -> Label`. Was that a purposeful change? It's nice not having to explicitly write the `Some(...)` wrapper around my strings, but now it looks like I need to again. I remember this being explicitly added in the past, so was this intentionally removed?

Susurrus:
Okay, started digging in to the git logs and it looks like this was a purposeful change due to the way you need to typecase `None`, which is annoying in many cases. So I'll update my code since the explicit `Option` looks to be necessary for the foreseeable future.

Susurrus:
Pushed a new version which *should* pass CI. So we'll see.

Susurrus:
Alright, it's all updated and passing in CI!

GuillaumeGomez:
Thanks! And yes it was on purpose. :)

PR #119, opened 2017-03-30
Add simple drag and drop example

Susurrus:
Still includes debugging output, this will be removed in final commit.

This is the same as #116 but pushes to the `pending` branch instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 target
+Cargo.lock
--- a/src/bin/drag_and_drop.rs
+++ b/src/bin/drag_and_drop.rs
@@ -1,0 +1,49 @@
+//! Simple drag and drop example
+//!
+//! Ported over from example code:
+//! https://developer.gnome.org/gtkmm-tutorial/stable/sec-dnd-example.html.en
+
+extern crate gdk;
+extern crate gtk;
+
+use gtk::prelude::*;
+
+fn main() {
+    if gtk::init().is_err() {
+        println!("Failed to initialize GTK.");
+        return;
+    }
+
+    // Configure button as drag source for text
+    let button = gtk::Button::new_with_label("Drag here");
+    let targets = vec![gtk::TargetEntry::new("STRING", gtk::TARGET_SAME_APP, 0),
+                       gtk::TargetEntry::new("text/plain", gtk::TARGET_SAME_APP, 0)];
+    button.drag_source_set(gdk::MODIFIER_MASK, &targets, gdk::ACTION_COPY);
+    button.connect_drag_data_get(|_, _, s, _, _| {
+                                     let data = "I'm data!";
+                                     s.set_text(data, data.len() as i32);
+                                 });
+
+    // Configure label as drag destination to receive text
+    let label = gtk::Label::new("Drop here");
+    label.drag_dest_set(gtk::DEST_DEFAULT_ALL, &targets, gdk::ACTION_COPY);
+    label.connect_drag_data_received(|w, _, _, _, s, _, _| { w.set_text(&s.get_text().unwrap()); });
+
+    // Stack the button and label horizontally
+    let hbox = gtk::Box::new(gtk::Orientation::Horizontal, 0);
+    hbox.pack_start(&button, true, true, 0);
+    hbox.pack_start(&label, true, true, 0);
+
+    // Finish populating the window and display everything
+    let window = gtk::Window::new(gtk::WindowType::Toplevel);
+    window.set_title("Simple Drag and Drop Example");
+    window.add(&hbox);
+    window.show_all();
+
+    // GTK & main window boilerplate
+    window.connect_delete_event(|_, _| {
+                                    gtk::main_quit();
+                                    Inhibit(false)
+                                });
+    gtk::main();
+}


src/drag_and_drop.rs
```diff
@@ -0,0 +1,73 @@
+//! Simple drag and drop example
+//!
+//! Ported over from example code:
+//! https://developer.gnome.org/gtkmm-tutorial/stable/sec-dnd-example.html.en
+extern crate gdk;
```
GuillaumeGomez: Can you add an empty line before this one please?

GuillaumeGomez:
Could you add more comments as well please?

Susurrus:
@GuillaumeGomez Are you referring to code comments or a better general description at the top?

GuillaumeGomez:
The top comment seems fine. I thought about explaining a bit more in the code directly.

Susurrus:
@GuillaumeGomez Will do.

GuillaumeGomez:
Thanks! :)

src/drag_and_drop.rs
```diff
@@ -0,0 +1,73 @@
...
+                        &targets,
+                        gdk::ACTION_COPY);
+    label.connect_drag_data_received(|w, d, _, _, s, _, t| {
+        println!("drag-data-received");
+        if s.get_length() > 0 && s.get_format() == 8 {
+            w.set_text(&s.get_text().unwrap());
+        }
+        d.drag_finish(false, false, t);
+    });
+    button.connect_drag_drop(|_, _, _, _, _| {
```
EPashkin: label.connect_drag_drop ;)

src/drag_and_drop.rs
```diff
@@ -0,0 +1,73 @@
...
+            println!("{}", t.name());
+        }
+    });
+    button.connect_drag_end(|_, _| {
+        println!("drag-end");
+    });
+
+    // Configure label as drag destination
+    let label = gtk::Label::new("Drop here");
+    label.drag_dest_set(gtk::DEST_DEFAULT_ALL,
```
EPashkin: If set gtk::DEST_DEFAULT_MOTION | gtk::DEST_DEFAULT_HIGHLIGHT and 
```
    label.connect_drag_drop(move |w, c, _, _, time| {
        println!("drag-drop");
        w.drag_get_data(c, &atom, time);
        true
    });
```
then `drag-data-received` called. May it be fixed with `Inhibit`

Susurrus: What's `&atom` here? This code doesn't compile.

EPashkin: Sorry, forgot add in begin main `let atom = gdk::Atom::intern("STRING");`

Susurrus:
I've pushed the most recent version that includes your comments here. I do see the same thing you're descripting, that `drag-data-received` isn't called when `DEST_DEFAULTS_ALL` is used. I checked the values for it, and they seem right (even though they seem to be computed in `gtk-sys` and then not re-exported in `gtk`, instead reimplemented), so I'm uncertain what the issue is.

In this version, though I don't see the data as being correctly received, even though the data seems set correctly in the `SelectionData` item during the `drag-data-get` callback.

EPashkin:
Yes, for unknown reason data from `data-get` don't accepted and `receive` called to inform about it.
This also reason why `receive` don't called with `DEST_DEFAULTS_ALL` as it call `receive` only on accepted data.
If we can get it work with current code, when we can restore `DEST_DEFAULTS_ALL` and remove `drag-drop`

Susurrus:
This behavior is weird because the call to `set_data` succeeds, which makes me think the error is somewhere else. I've also added another example I want to add at some point where you can drag a file onto a `TextView` and it'll show the contents in it. This one uses the "default" configuration of a drag from outside the application, and the `drag-data-received` signal is called correctly with `DEST_DEFAULT_ALL` set. So I'm thinking there's something specific to the `drag-data-get` signal that is causing the problem.

**Note:** there are various other bugs that are apparent, either in my code or in gtk-rs when using the drag_and_drop_textview example, so it'll likely crash on you as you play around with it.

I'm going to reimplement this simple drag-and-drop example in C and see if I can get it to work correctly there so we at least have a sound reference. This drag and drop stuff is pretty confusing, so I'm not even sure if I'm trying to use it correctly.

Susurrus:
So I worked on a simple C implementation to verify that I knew enough what I was doing to actually understand the Rust version. See below:
```C
//gcc -Wall drag_and_drop.c -o drag_and_drop `pkg-config --cflags --libs gtk+-3.0`

#include <stdio.h>
#include <gtk/gtk.h>
#include <gdk/gdk.h>

static void drag_data_get(GtkWidget *widget, GdkDragContext *d, GtkSelectionData *s, guint32 x, guint32 y)
  {
    printf("drag-data-get\n");
    gtk_selection_data_set_text(s, "I'm data!", -1);
  }

static void drag_data_received(GtkWidget *widget, GdkDragContext *d, gint32 i, gint32 j, GtkSelectionData *s, guint32 x, guint32 y)
  {
    printf("drag-data-received\n");
    guchar *text = gtk_selection_data_get_text(s);
    gtk_label_set_text(GTK_LABEL(widget), text);
  }

static gboolean drag_drop(GtkWidget *widget, GdkDragContext *d, gint x, gint y, guint time)
  {
    printf("drag-drop\n");
    GdkAtom target = gdk_atom_intern("STRING", FALSE);
    gtk_drag_get_data(widget, d, target, time);
    return TRUE;
  }

int main(int argc, char **argv)
  {     
    gtk_init(&argc, &argv);   


    GtkWidget *button = gtk_button_new_with_label("Drag here");
    GtkTargetEntry targets[] = {
        {"STRING", 0, 0},
	{"text/plain", 0, 0},
    };
    gtk_drag_source_set(button, GDK_MODIFIER_MASK, targets, 2, GDK_ACTION_COPY);
    g_signal_connect(button, "drag-data-get", G_CALLBACK(drag_data_get), NULL);
    GtkWidget *window=gtk_window_new(GTK_WINDOW_TOPLEVEL);
    g_signal_connect(window, "destroy", G_CALLBACK(gtk_main_quit), NULL);
       
    GtkWidget *label = gtk_label_new("Drop here");
    gtk_drag_dest_set(label, GTK_DEST_DEFAULT_ALL, targets, 2, GDK_ACTION_COPY);
    g_signal_connect(label, "drag-data-received", G_CALLBACK(drag_data_received), NULL);
    //g_signal_connect(label, "drag-drop", G_CALLBACK(drag_drop), NULL);
     
    GtkWidget *hbox = gtk_box_new(GTK_ORIENTATION_HORIZONTAL, 0);
    gtk_box_pack_start(GTK_BOX(hbox), button, TRUE, TRUE, 0);
    gtk_box_pack_start(GTK_BOX(hbox), label, TRUE, TRUE, 0);

    gtk_container_add(GTK_CONTAINER(window), hbox);
   
    gtk_widget_show_all(window);                 
    gtk_main();

    return 0;
  }
```

Now when playing around with this code a bit, something interesting happens. For the C-code example, you can drag the button onto a terminal or other window and the `"I'm data!"` text will show up. Additionally, if you run both programs you can drag from the C-example button into the Rust-example label and the text will show up properly. But nothing happens going the opposite direction. I agree with your assessment @EPashkin that the bug here is with `SelectionData` somehow and not with the drag-and-drop implementations.

Susurrus:
These examples should both work correctly with gtk-rs/gtk#472 applied. I've only tested `drag_and_drop_textview` on Linux, but now I'm using `url` which will hopefully address cross-platform issues. I think @EPashkin you're on Windows, so let me know if you have file path problems again.

EPashkin:
I like `drag_and_drop.rs` but `drag_and_drop_textview.rs` IMHO unneeded.
It too error prone, works bad with not utf8 files and fixing last need code that don't show anything about D&D.
Only difference from base is "text/uri-list".
cc @GuillaumeGomez

EPashkin:
Url fixed problem with pathes but showing file is still problem

src/drag_and_drop_textview.rs
```diff
@@ -0,0 +1,77 @@
+//! More complex drag and drop example
+//!
+//! Displays the text of a file in a TextView widget after it has been dragged
+//! and dropped onto it.
+//! http://www.kksou.com/php-gtk2/sample-codes/display-a-text-file-in-GtkTextView-with-drag-and-drop.php
+extern crate gdk;
```
GuillaumeGomez: Please put an empty line here.

Susurrus:
One of the reasons for the two different examples is that the simple one is between a single application (I need to re-add the appropriate source filter, I removed it at some point) and the second shows both how to deal with URI lists and also supporting drops from an external application. If the file reading is too complication, I can just display a bulleted list in the textview with the full names of all of the files dropped on it. One of the reasons I set up the second example is because I used it as a minimal test case for my application where I wanted to received file names from drags. So I used that example to test.

EPashkin:
IMHO file list will better.
Also I forgot to say that on windows d&d example don't show mouse pointer when dragging,
on linux it shown?

Susurrus:
I've modified things as we've discussed and I also use `rustfmt` to clean up the code a tad.

@EPashkin If you'd like to run the `drag_and_drop_textview` example, I'm getting crashes using `TextBuffer::InsertAtCursor`, which are interesting. Even if you change the line to just insert static text like `buffer.insert_at_cursor("Hey!")`, I still get crashes (not panics). Can you confirm that you see these on your end as well?

EPashkin:
Both cases normal for me. But it drop different selection it changed only on second drop (it changed on second drop even if I drop first selection

EPashkin:
When I add count to output: i get 1 then 3 on second drop, then 5 ...

Susurrus:
Hmm, I'm running into multiple crashes and other failures using this example that I don't think should be happening:

If I drag a single file onto the TextView, it's displayed fine. If I then drag a different file onto the window in a second drag-and-drop event, the program crashes. Sometimes I get a `drag_and_drop_textview: cairo-region.c:401: cairo_region_reference: Assertion 'CAIRO_REFERENCE_COUNT_HAS_REFERENCE (&region->ref_count)' failed.` error.

And if I drag multiple folders onto the window it will pretty consistently crash with no warning output.

I'm also seeing that the iteration order is backwards from what I'd expect for the `Vec` returned from `SelectionData::get_uris()`. I think that's what you're referring to in your last comment? This might actually be the problem here. I've pushed a new version that has some `println` macros to output debugging info and I think it revealed that the problem is in `get_uris`. Running it with a few files of mine I see the following:
```
a - ["file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-06.png", "file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-14.png", "file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-15.png"]
b - "file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-15.png"
c - "/home/susurrus/Screenshot from 2017-02-09 18-25-15.png"
d - "/home/susurrus/Screenshot from 2017-02-09 18-25-15.png"
e - " • /home/susurrus/Screenshot from 2017-02-09 18-25-15.png\n"
b - "file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-06.png"
c - "/home/susurrus/Screenshot from 2017-02-09 18-25-06.png"
d - "/home/susurrus/Screenshot from 2017-02-09 18-25-06.png"
e - " • /home/susurrus/Screenshot from 2017-02-09 18-25-06.png\n"
b - "file:///home/susurrus/Screenshot%20from%202017-02-09%2018-25-06.png"
c - "/home/susurrus/Screenshot from 2017-02-09 18-25-06.png"
d - "/home/susurrus/Screenshot from 2017-02-09 18-25-06.png"
e - " • /home/susurrus/Screenshot from 2017-02-09 18-25-06.png\n"
```

Notice that the iteration order is actually 3, 1, 1 instead of the 1, 2, 3 we'd expect. I'm wondering if this bug is within `SelectionData` itself. It looks like string pointers are messed up somewhere when multiple uris are found.

P.S. I also added `Cargo.lock` to `.gitignore` because it was bothering me it showing up in `git status`.

EPashkin:
Sorry, I currently don't have time to look (looked for cairo fixes)
Reversed order come from realization of `ToGlibPtr` for [GList](https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L486-L535). May be this need be fixed with `iter().reverse()`

src/bin/drag_and_drop_textview.rs
```diff
@@ -0,0 +1,66 @@
...
+    let scrolled_text_view = gtk::ScrolledWindow::new(None, None);
+    scrolled_text_view.set_policy(gtk::PolicyType::Automatic, gtk::PolicyType::Automatic);
+    scrolled_text_view.add(&text_view);
+    let targets = vec![gtk::TargetEntry::new("text/uri-list", gtk::TARGET_OTHER_APP, 0)];
+    text_view.drag_dest_set(gtk::DEST_DEFAULT_ALL, &targets, gdk::ACTION_COPY);
+    text_view.connect_drag_data_received(|w, _, _, _, d, _, _| {
+        let buffer = w.get_buffer().unwrap();
+        buffer.set_text("");
+        println!("a - {:?}", d.get_uris());
+        for file in d.get_uris() {
```
EPashkin: Move result `get_uris` to variable

EPashkin:
@Susurrus 
Seems I found problem with crashes on Ubuntu: https://github.com/gtk-rs/glib/pull/160
Unsolved problems with drag_and_drop_textview.rs:
On Ubuntu called 2 times for one drop. Even `context.drag_finish(true, false, time_);` don't fix it.
On Windows it even more strange:
drop file1: `b - file1`
drop file2:
```
b - file2
b - file1
```
drop file3:
```
b - file3
b - file2
```
Also drag_and_drop.rs on Windows don't shows any icon on dragging (even no mouse pointer)

EPashkin:
For fix double call we can use this hack:
```rust
let last_received_time = RefCell::new(0);
text_view.connect_drag_data_received(move |w, c, _, _, d, _, time_| {
    if time_ <= *last_received_time.borrow() { return; }
    *last_received_time.borrow_mut() = time_;
```

Susurrus:
I don't see the multiple drops at a time. What version of gtk are you using on Ubuntu, maybe it's an old bug that's been fixed for me?

I'll try to implement the complex example in C as well to see how it should perform on my version and that should help in debugging.

EPashkin:
I have libgtk-3-dev:amd64    3.18.9-1ubuntu3.3 installed
```terminal
$ uname -a
Linux Ubuntu 4.4.0-75-generic #96-Ubuntu SMP Thu Apr 20 09:56:33 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 16.04.2 LTS
Release:	16.04
Codename:	xenial
```

EPashkin:
.. it installed on virtualbox.

As for https://github.com/gtk-rs/glib/pull/160#issuecomment-298796934 I don't see any warning.

Susurrus:
Yeah, that's a pretty old version of GTK; I'm running 3.22.12. Let's do this. I'll pare this PR down to just the simple DnD example and open a new PR for the file list one where we can continue this. I'll also make a C version for that so we can see if it works under the proper situation before trying to debug the gtk-rs problems on top of that.

Susurrus:
Alright, I've reduced this down to just the simple drag and drop example that has no issues with it and should be clear to merge. I'll get working on the other example and send up a new PR for it.

EPashkin:
There still one problem last with drag_and_drop.rs too: on windows + msys2 it don't show any mouse pointer on dragging.

Susurrus:
There's no mouse configuration within this functionality. I don't set a mouse icon, it does whatever the default GTK+ stuff does. So this is either a GTK+ bug or maybe you don't have the icon set installed for GTK+ so it's missing that and rendering nothing instead.

EPashkin:
Yes, seems so.

Susurrus:
@EPashkin Is this good to merge?

EPashkin:
@Susurrus IMHO - yes

GuillaumeGomez:
Thanks!

PR #121, opened 2017-04-06
Add multi window example

GuillaumeGomez:
Fixes gtk-rs/gtk/issues/484.

### DIFF
--- a/src/bin/multi_windows.rs
+++ b/src/bin/multi_windows.rs
@@ -1,0 +1,114 @@
+extern crate gtk;
+
+use gtk::prelude::*;
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+use std::rc::Rc;
+
+fn create_sub_window(title: &str, main_window_entry: gtk::Entry, id: usize,
+                     windows: Rc<RefCell<HashMap<usize, gtk::Window>>>) {
+    let window = gtk::Window::new(gtk::WindowType::Toplevel);
+
+    window.set_title(title);
+    window.set_default_size(400, 200);
+
+    let windows_clone = windows.clone();
+    window.connect_delete_event(move |_, _| {
+        windows_clone.borrow_mut().remove(&id);
+        Inhibit(false)
+    });
+
+    let button = gtk::Button::new_with_label(&format!("Notify main window with id {}!", id));
+    button.connect_clicked(move |_| {
+        // When the button is clicked, let's write it on the main window's entry!
+        main_window_entry.get_buffer().set_text(&format!("sub window {} clicked", id));
+    });
+    window.add(&button);
+
+    window.show_all();
+    // Once the new window has been created, we put it into our hashmap so we can update its
+    // title when needed.
+    windows.borrow_mut().insert(id, window.clone());
+}
+
+fn create_main_window() -> gtk::Window {
+    let window = gtk::Window::new(gtk::WindowType::Toplevel);
+
+    window.set_title("I'm the main window");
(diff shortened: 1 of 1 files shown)


src/multi_windows.rs
```diff
@@ -0,0 +1,116 @@
...
+    });
+
+    window.show_all();
+    window
+}
+
+fn generate_new_id(windows: Rc<RefCell<HashMap<usize, gtk::Window>>>) -> usize {
+    let mut len = {
+        windows.borrow().values().count()
+    };
```
EPashkin: IMHO unneeded {}

EPashkin:
Recreate with pending branch please

src/multi_windows.rs
GuillaumeGomez: Actually no, it's not. Good point. Just setting to 0 is fine.

EPashkin: If you set 0, then will be unneeded loop.

EPashkin:
I don't know that editing can change branch.
@GuillaumeGomez, can you also change others PRs?

src/multi_windows.rs
GuillaumeGomez: The loop is still needed because I need every subwindow to have a different id.

GuillaumeGomez:
Dark magic! :p

And sure.

EPashkin:
May be we really need change branch setup for this crate to master/production

GuillaumeGomez:
Instead of doing this, I think we should maybe add a README note about PRs or maybe a PR message (since github allows it...).

EPashkin:
It already added https://github.com/gtk-rs/examples/blob/master/CONTRIBUTING.md and shown only when attempt to add PR on master. Just not effective.

GuillaumeGomez:
Well, we tried. :p

GuillaumeGomez:
@EPashkin: Updated.

EPashkin:
@GuillaumeGomez file not in Bin

GuillaumeGomez:
Updated.

src/bin/multi_windows.rs
```diff
@@ -0,0 +1,114 @@
...
+        gtk::main_quit();
+        Inhibit(false)
+    });
+
+    window.show_all();
+    window
+}
+
+fn generate_new_id(windows: Rc<RefCell<HashMap<usize, gtk::Window>>>) -> usize {
+    let mut len = 0;
```
EPashkin: Maybe it `id` now?

GuillaumeGomez: Fixed.

src/bin/multi_windows.rs
```diff
@@ -0,0 +1,114 @@
...
+    let button = gtk::Button::new_with_label("Create new window");
+    button.connect_clicked(move |_| {
+        let new_id = generate_new_id(windows.clone());
+        let new_window = create_sub_window(&windows_title_entry_clone.get_buffer().get_text(),
+                                           main_entry.clone(),
+                                           new_id,
+                                           windows.clone());
+        // Once the new window has been created, we put it into our hashmap so we can update its
+        // title when needed.
+        windows.borrow_mut().insert(new_id, new_window);
```
EPashkin: May be this better go to end of `create_sub_window` this way all logic will be almost in one place

GuillaumeGomez: Hum, good idea.

GuillaumeGomez: Updated.

EPashkin:
Thanks

PR #123, opened 2017-04-15
Moved all examples to `bin` subfolder

EPashkin:
Fixes build error on cargo 0.19.0-nightly (c416fb60b 2017-04-05) and any future [[bin]] removal

PR #128, opened 2017-04-19
Moved all examples to `bin` subfolder

EPashkin:


Susurrus:
There is actually support in cargo for `examples`. You can run them like `caro run --example XXX`. I think that only makes more sense than this if these are moved into the main gtk repo, however.

EPashkin:
--examples IMHO works for files in `examples` and it not added to toml normally.
This crate contains only examples and its runs with --bin.
Latest nightly only works with binaries in `bin` folder.

Susurrus:
Yes, my suggestion was to move everything to an `examples` folder and still remove them all from `Cargo.toml`. Then things can be run by calling `cargo run --example XXX` instead of `cargo run --bin XXX`. What's nice about examples is that when running `cargo test` it will compile-test all examples. I don't believe that happens for binaries. This would give some good ergonomics to the `gtk-rs/gtk` library if these examples are merged there because then `cargo test` will make sure all examples still compile.

EPashkin:
--examples IMHO used only when other part present too (lib or bin).
As we build examples on CI bulds in all your crates it has separate crate (maybe replaced with https://github.com/gtk-rs/tests if there too many examples here to minimize build time),
no difference between direct build and build on test

Susurrus:
I didn't even realize tests were separate. I would think it'd be easier to have both in the main gtk-rs/gtk repo, but I'm assuming these separate repos exist for a reason.

I'm +1 on getting them out of Cargo.toml, as modifying that was annoying to add new examples.

EPashkin:
Ping

GuillaumeGomez:
Thanks!

Susurrus:
Is this going to be merged into the `pending` branch as well?

EPashkin:
`pending` was first with #123

Susurrus:
@EPashkin Sorry! Forgot to fetch the latest from upstream. Apologies for the noise.

PR #133, opened 2017-06-05
Use `Result::expect` instead of checking `is_err`

Wallacoloo:
This is generally more idiomatic rust in addition to being more concise.

I'm a bit confused as per the meaning of `CONTRIBUTING.md`. This change set targets all examples in master, and some of those examples aren't present in the `pending` branch. But the document very clearly says not to merge against master. I could pare this down to just target the examples that are present in `pending` to allow auto-merge if desired, but I don't see the point.


EPashkin:
Please make it from pending branch: too many bad changes

Wallacoloo:
Ok. I created a new PR for that: #134.

PR #134, opened 2017-06-05
Use `Result::expect` instead of checking `is_err`

Wallacoloo:
Identical to #133 but only fixes the examples in the `pending` branch.

### DIFF
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -9,10 +9,7 @@ extern crate gtk;
 use gtk::prelude::*;
 
 fn main() {
-    if gtk::init().is_err() {
-        println!("Failed to initialize GTK.");
-        return;
-    }
+    gtk::init().expect("Failed to initialize GTK.");
 
     let window = gtk::Window::new(gtk::WindowType::Toplevel);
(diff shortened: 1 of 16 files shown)


GuillaumeGomez:
The problem with this solution is that it panics and makes an ugly output, so I don't like much. :-/

What do you think @EPashkin?

EPashkin:
@Wallacoloo, Sorry I don't looked what really changed in previous PR.
I agree with @GuillaumeGomez that `expect` produce output, that unneeded in this case.
PS. All example targets is equals for both master and pending.

Wallacoloo:
When it comes to writing example code, my opinion is that having idiomatic code is more important than pretty output (though I also don't find the output of panics to be ugly). Panics also point you directly to the source of the error (line number / source file), so again I think that's a useful thing to provide developers.

And for what it's worth, [one of the examples](https://github.com/gtk-rs/examples/blob/master/src/bin/child-properties.rs#L13) already panics on error.

This is all just opinions though - do as you see fit.

GuillaumeGomez:

PR #136, opened 2017-07-08
Add example for pango text attributes

EPashkin:
cc @GuillaumeGomez

### DIFF
--- a/src/bin/pango_attributes.rs
+++ b/src/bin/pango_attributes.rs
@@ -1,0 +1,56 @@
+//! # Pango text attributes
+//!
+//! This sample demonstrates how to use various attributes on labels text.
+
+#![crate_type = "bin"]
+
+extern crate gtk;
+extern crate pango;
+
+use gtk::prelude::*;
+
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Please add explanations about what the code is doing (it's examples after all). :)

EPashkin:
Updated

src/bin/pango_attributes.rs
```diff
@@ -0,0 +1,56 @@
+//! # Pango text attributes
+//!
+//! This sample demonstrates how to use various attributes (strikeout, background color, etc. ) on labels text
```
GuillaumeGomez: Missing dot at the end of the line. Also, can you keep the line's length under 100 column please?

GuillaumeGomez:
Not much comments, but considering the "complexity" of the example, I think just the description at the beginning of the file is enough.

EPashkin:
Updated

GuillaumeGomez:
Thanks!

PR #147, opened 2017-08-31
Get latest rust version via rustup

EPashkin:
Same as https://github.com/gtk-rs/sourceview/pull/16

cc @GuillaumeGomez


GuillaumeGomez:
Thanks!

PR #150, opened 2017-09-23
Application

GuillaumeGomez:
cc @gkoz @sdroege

### DIFF
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -1,35 +1,64 @@
 //! # Basic Sample
 //!
-//! This sample demonstrates how to create a toplevel `window`, set its title, size and position, how to add a `button` to this `window` and how to connect signals with actions.
+//! This sample demonstrates how to create a toplevel `window`, set its title, size and
+//! position, how to add a `button` to this `window` and how to connect signals with
+//! actions.
 
-#![crate_type = "bin"]
-
+extern crate gio;
 extern crate gtk;
 
+use gio::prelude::*;
 use gtk::prelude::*;
 
-fn main() {
-    if gtk::init().is_err() {
-        println!("Failed to initialize GTK.");
-        return;
-    }
+use std::env::args;
 
-    let window = gtk::Window::new(gtk::WindowType::Toplevel);
+// make moving clones into closures more convenient
+macro_rules! clone {
+    (@param _) => ( _ );
+    (@param $x:ident) => ( $x );
+    ($($n:ident),+ => move || $body:expr) => (
+        {
+            $( let $n = $n.clone(); )+
+            move || $body
+        }
+    );
+    ($($n:ident),+ => move |$($p:tt),+| $body:expr) => (
+        {
+            $( let $n = $n.clone(); )+
+            move |$(clone!(@param $p),)+| $body
+        }
+    );
+}
+
+fn build_ui(application: &gtk::Application) {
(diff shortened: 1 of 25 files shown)


src/bin/basic.rs
```diff
@@ -1,35 +1,64 @@
...
-}
\ No newline at end of file
+}
+
+fn main() {
+    let application = gtk::Application::new("com.github.basic",
+                                            gio::ApplicationFlags::empty())
+                                       .expect("Initialization failed...");
+
+    application.connect_startup(move |app| {
```
EPashkin: `move` unneeded here

EPashkin:
Only  problem with be that on windows, that it created listeners for `tcp://[::1] Port 54038` (port changes)
by running `"C:\WINDOWS\system32\rundll32.exe" D:\P\msys64\mingw64\bin\LIBGIO~1.DLL,g_win32_run_session_bus`.
If firewall disallow create listeners, then program stall for ~40sec.

src/bin/basic.rs
```diff
@@ -1,35 +1,64 @@
...
 
-fn main() {
-    if gtk::init().is_err() {
-        println!("Failed to initialize GTK.");
-        return;
-    }
+use std::env::args;
+
+// make moving clones into closures more convenient
+macro_rules! clone {
```
EPashkin: Seems we need add this macro to glib, or use separate crate (I seen it in this month but can't remember when).

EPashkin:
Also maybe this part better added to `gtk::Window` as `connect_application_and_show_all` ?
Of course without usage of `clone!`.
```
        window.set_application(application);
        window.connect_delete_event(clone!(window => move |_, _| {
            window.destroy();
            Inhibit(false)
        }));
        window.show_all();
```
PS. `window.set_application(application);` not present in `basic.rs` or some others.

sdroege:
> Only problem with be that on windows, that it created listeners for tcp://[::1] Port 54038 (port changes)
by running "C:\WINDOWS\system32\rundll32.exe" D:\P\msys64\mingw64\bin\LIBGIO~1.DLL,g_win32_run_session_bus.
If firewall disallow create listeners, then program stall for ~40sec.

Please file a bug against GTK for that, that seems unacceptable

src/bin/basic.rs
sdroege: https://crates.io/crates/closet

sdroege:
> Also maybe this part better added to gtk::Window as connect_application_and_show_all ?

While it's used like this often in the example, it could very well be different in real applications.

src/bin/basic.rs
EPashkin: Yes, thanks.

EPashkin:
@sdroege you right about `connect_application_and_show_all`. Normal application may need do some work on exiting.


About listener:
it caused by nonexistence build-in DBus for Windows.
If any example running, others don't create new listeners and use first's.
So IMHO nothing to report.

src/bin/basic.rs
GuillaumeGomez: I was just thinking adding a new crate to `gtk-rs` with some utils like that.

GuillaumeGomez:
> PS. `window.set_application(application);` not present in `basic.rs` or some others.

`set_application` is useful only for glade examples in order to set the setup `application` to the `ApplicationWindow`. Otherwise it's useless.

GuillaumeGomez:
I'm merging it then.

PR #156, opened 2017-11-23
Follow cargo standard directory structure for examples

creativcoder:
Current directory structure, does not allow running examples directly with cargo like:
`cargo run --example multi_windows`

This PR organizes directory structure as mentioned in [Cargo manifest example section](http://doc.crates.io/manifest.html#examples) and allows running examples directly from cargo, without editing of Cargo.toml file

Readme is updated to show how to run the examples.

Let me know if this PR missed something


GuillaumeGomez:
Because it's supposed to be run with `cargo run --bin multi_windows`. Also, we don't accept PRs on master. Thanks though!

PR #159, opened 2018-01-22
Configure parent window for dialog in builder_basics.glade.

tmiasko:


tmiasko:
Yes, it avoids warning messages. From more functional perspective it hints to
the window manager to place the dialog above the main window. On my system, the
main window also happens to be grayed out to indicate that it is inactive until
dialog is closed.

GuillaumeGomez:
Thanks!

PR #160, opened 2018-01-22
Remove redundant delete event handler.

tmiasko:
Commit that introduced gtk::Application also modified window delete
event handlers, replacing calls to gtk::main_quit with ones that destroy
the application window.

The new event handler is redundant as the default delete event handler
already destroys the window.

### DIFF
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -12,23 +12,6 @@ use gtk::prelude::*;
 
 use std::env::args;
 
-// make moving clones into closures more convenient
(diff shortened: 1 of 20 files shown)


tmiasko:
If I understand you correctly, this means that CMD-Q emits delete event, but
does not destroy the window by default? That's really surprising and unfortunate.

Oh, well. In this case feel free to close this PR if you want to preserve existing behaviour.

GuillaumeGomez:
I'd prefer to. Thanks anyway!

Issue #162, opened 2018-04-06
gtktest consumes ~75% of cpu on OS X


ehiggs:
Reproduction: 
Build examples using `cargo build --release --features gtk_3_10` on OS X El Capitan.

Observed:
~75% cpu usage (and 45M resident memory use) as seen through `top -o cpu`

Expected: 
As an example, gtktest should consume almost no resources as it is not doing anything afaict.

EPashkin:
Seems it because `GtkSpinner` implementation as it working all time,
without it gtktest don't consume CPU.

ehiggs:
75% cpu seems excessive for drawing a spinner. Is this a gtk bug or gtk-rs?

EPashkin:
Gtk's, as gtk-rs only wrapper.
There was already many bugs filled in Gtk like https://bugzilla.gnome.org/show_bug.cgi?id=684639

PR #163, opened 2018-04-06
Disable spinner in glade file, to make starting it in code meaningful

EPashkin:
Fix double activating found thanks to https://github.com/gtk-rs/examples/issues/162

We already have activating in code:
```rust
let spinner: Spinner = builder.get_object("spinner").expect("Couldn't get spinner");
spinner.start();
```

cc @GuillaumeGomez


GuillaumeGomez:
:+1: (just waiting for CIs)

PR #166, opened 2018-04-20
Add gio_futures example

sdroege:
Depends on https://github.com/gtk-rs/gio/pull/108

I'd also like to add an example using async/await but that would require nightly. How would we handle that?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,10 @@ authors = ["The Gtk-rs Project Developers"]
 [dependencies]
 chrono = "0.4"
 url = "1.4"
+futures-core = { version = "0.2", optional = true }
+futures-executor = { version = "0.2", optional = true }
+futures-channel = { version = "0.2", optional = true }
+futures-util = { version = "0.2", optional = true }
 
 [dependencies.glib]
 git = "https://github.com/gtk-rs/glib"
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
By adding a feature called "nighty-only" or something along the lines? I restarted CIs.

Cargo.toml
```diff
@@ -36,3 +40,80 @@ gtk_3_16 = ["gtk_3_10", "gtk/v3_16"]
 gtk_3_18 = ["gtk_3_16", "gtk/v3_18"] #for CI tools
 gtk_3_20 = ["gtk_3_18", "gtk/v3_20"] #for CI tools
 gtk_3_22 = ["gtk_3_20", "gtk/v3_22"] #for CI tools
+futures = ["futures-core", "futures-executor", "futures-channel", "futures-util", "glib/futures", "gio/futures"]
+
+[patch.'https://github.com/gtk-rs/gio']
+gio = { git = 'https://github.com/sdroege/gio-rs', branch = 'futures' }
```
sdroege: This currently patches in my GIO branch from https://github.com/gtk-rs/gio/pull/108. Will remove once that is merged

Cargo.toml
```diff
@@ -36,3 +40,80 @@ gtk_3_16 = ["gtk_3_10", "gtk/v3_16"]
...
+[[bin]]
+name = "text_viewer"
+
+[[bin]]
+name = "transparent_main_window"
+
+[[bin]]
+name = "treeview"
+
+[[bin]]
```
EPashkin: Strange, why [[bin]] added?

EPashkin: Oh, understand

EPashkin: Maybe better do same as we do with version?
https://github.com/gtk-rs/examples/blob/master/src/bin/builder_basics.rs#L76-L85

Cargo.toml
```diff
@@ -36,3 +40,80 @@ gtk_3_16 = ["gtk_3_10", "gtk/v3_16"]
 gtk_3_18 = ["gtk_3_16", "gtk/v3_18"] #for CI tools
 gtk_3_20 = ["gtk_3_18", "gtk/v3_20"] #for CI tools
 gtk_3_22 = ["gtk_3_20", "gtk/v3_22"] #for CI tools
+futures = ["futures-core", "futures-executor", "futures-channel", "futures-util", "glib/futures", "gio/futures"]
```
EPashkin: Maybe here also must be some `gtk_3_xxx` ?

EPashkin:
And IMHO https://github.com/gtk-rs/examples/blob/pending/build_travis.sh#L8 IMHO also need `futures`

Cargo.toml
sdroege: No GTK is needed here

sdroege:
> And IMHO https://github.com/gtk-rs/examples/blob/pending/build_travis.sh#L8 also need futures

Agreed

Cargo.toml
sdroege: Why do in code what cargo can do for us? :) I would rather convert all the examples

Cargo.toml
EPashkin: Ok

Cargo.toml
EPashkin: Current version don't accepted by cargo as it don't have `path`.

EPashkin:
Please, add futures to `#default = ["gtk_3_22"]` too, it helps greatly for local checking

Cargo.toml
EPashkin: `"glib/futures"` IMHO unneeded here, as it included in `"gio/futures"`

Cargo.toml
EPashkin: Oops, it just because `variant.rs` do not exists

sdroege: Hmm I have that file :) I'll check and fix later

sdroege:
async-await example would also require using `futures-stable`, `Pin` and all that inside glib for the `Executor` and everything. I'd postpone that for some other time

sdroege:
> async-await example would also require using futures-stable, Pin and all that inside glib for the Executor and everything. I'd postpone that for some other time

And with that I mean in the next days in separate PRs, not the too distant future. I have something half-working

sdroege:
@GuillaumeGomez Let's get this in then, or is there anything missing still?

GuillaumeGomez:
I was waiting for CIs and forgot about it. Sorry about that...

PR #172, opened 2018-06-27
Use futures-preview crates instead of futures

sdroege:
The futures crates were janked from crates.io and renamed to
futures-preview for whatever reason.

Nightly still does not compile because `futures-core-preview` does not compile


GuillaumeGomez:
Thanks!

EPashkin:
@sdroege You probably meant "futures-nightly" feature not works, as CI seems works fine 😉 
https://ci.appveyor.com/project/GuillaumeGomez/examples/build/1.0.395
https://travis-ci.org/gtk-rs/examples/builds/397266103

PR #175, opened 2018-07-10
Add GtkListBox example with model, property bindings and a custom row…

sdroege:
… GObject type

### DIFF
--- a/src/bin/listbox_model.rs
+++ b/src/bin/listbox_model.rs
@@ -1,0 +1,322 @@
+//! # ListBox and ListModel Sample
+//!
+//! This sample demonstrates how to use gtk::Listbox in combination with
+//! gio::ListStore as a model with a custom row type.
+
+#[macro_use]
+extern crate glib;
+extern crate gio;
+extern crate gtk;
+
+extern crate glib_sys as glib_ffi;
+extern crate gobject_sys as gobject_ffi;
+
+extern crate gobject_subclass;
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+use std::env::args;
+
+use row_data::RowData;
+
+// make moving clones into closures more convenient
+macro_rules! clone {
+    (@param _) => ( _ );
+    (@param $x:ident) => ( $x );
+    ($($n:ident),+ => move || $body:expr) => (
+        {
+            $( let $n = $n.clone(); )+
+            move || $body
+        }
+    );
+    ($($n:ident),+ => move |$($p:tt),+| $body:expr) => (
+        {
+            $( let $n = $n.clone(); )+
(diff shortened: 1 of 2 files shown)


EPashkin:
@sdroege Big thanks, 👍

src/bin/listbox_model.rs
```diff
@@ -0,0 +1,322 @@
...
+            dialog.connect_response(|dialog, _| dialog.destroy());
+
+            let content_area = dialog.get_content_area();
+
+            let entry = gtk::Entry::new();
+            item.bind_property("name", &entry, "text")
+                .flags(glib::BindingFlags::DEFAULT | glib::BindingFlags::SYNC_CREATE | glib::BindingFlags::BIDIRECTIONAL)
+                .build();
+            entry.connect_activate(clone!(dialog => move |_| {
+                dialog.response(0);
```
EPashkin: @sdroege What does this signal handler do?

sdroege: When you press enter inside the `GtkEntry`, it sends response `0` (see above, "Ok") to the dialog. It is as if you clicked the corresponding button on the dialog

EPashkin: Thanks

sdroege:
And all green :)

EPashkin:
cc @GuillaumeGomez

src/bin/listbox_model.rs
```diff
@@ -0,0 +1,322 @@
...
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+use std::env::args;
+
+use row_data::RowData;
+
+// make moving clones into closures more convenient
+macro_rules! clone {
```
GuillaumeGomez: We should really add this into a crate or somewhere...

GuillaumeGomez:
Can you please add a few comments along the code? It'd help readers to understand what's happening. :)

src/bin/listbox_model.rs
sdroege: clone_army! in the closet crate :)

sdroege:
Sure, will do :) are there any specifics parts that you would like to have explained in more details?

Apart from comments it looks good to you?

src/bin/listbox_model.rs
GuillaumeGomez: XD

GuillaumeGomez:
Yes, just few comments. And explain what's going on in some cases that you think could be tricky. For us, gnome users, it's hard to know which parts are difficult to understand to newcomers so I can't help you much on that. :-/

sdroege:
Added lots of comments, no code changes

GuillaumeGomez:
Thanks a lot! :)

PR #176, opened 2018-07-12
Make use of the new downcast_ref()

sdroege:


EPashkin:
Can it be used in gtktest.rs too?

EPashkin:
Bad.
Then be only one change.
Thanks and 👍

GuillaumeGomez:
Thanks!

PR #177, opened 2018-07-13
Add accelerators to menu examples

FreeFull:
"Accelerator" is the term GTK uses for keyboard shortcuts such as `Ctrl+Q`. The shortcuts get displayed for items in a menu, which seems to make these two examples the best for also demonstrating accelerators.

### DIFF
--- a/src/bin/menu_bar.rs
+++ b/src/bin/menu_bar.rs
@@ -10,8 +10,8 @@ extern crate gtk;
 use gio::{ApplicationExt, ApplicationExtManual};
 use gtk::prelude::*;
 use gtk::{
-    AboutDialog, ApplicationWindow, CheckMenuItem, IconSize, Image, Label, Menu, MenuBar, MenuItem,
-    WindowPosition,
+    AboutDialog, AccelFlags, AccelGroup, ApplicationWindow, CheckMenuItem, IconSize, Image, Label,
+    Menu, MenuBar, MenuItem, WindowPosition,
 };
 
 use std::env::args;
@@ -49,6 +49,8 @@ fn build_ui(application: &gtk::Application) {
     let v_box = gtk::Box::new(gtk::Orientation::Vertical, 10);
 
     let menu = Menu::new();
+    let accel_group = AccelGroup::new();
+    window.add_accel_group(&accel_group);
     let menu_bar = MenuBar::new();
     let file = MenuItem::new_with_label("File");
     let about = MenuItem::new_with_label("About");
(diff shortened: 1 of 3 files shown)


src/bin/menu_bar.rs
```diff
@@ -97,6 +102,14 @@ fn build_ui(application: &gtk::Application) {
         window.destroy();
     }));
 
+    quit.add_accelerator(
```
GuillaumeGomez: This might not seem easy to understand for people who don't know GTK. Can you add a comment explaining that this control is expecting a "CTRL+Q" or a "CMD+Q" please?

src/bin/menu_bar.rs
```diff
@@ -49,6 +51,9 @@ fn build_ui(application: &gtk::Application) {
     let v_box = gtk::Box::new(gtk::Orientation::Vertical, 10);
 
     let menu = Menu::new();
+    let accel_group = AccelGroup::new();
+    window.add_accel_group(&accel_group);
```
GuillaumeGomez: Might be useful to explain why you added the accel group on both `window` and `menu`.

src/bin/menu_bar.rs
```diff
@@ -97,6 +102,14 @@ fn build_ui(application: &gtk::Application) {
         window.destroy();
     }));
 
+    quit.add_accelerator(
+        "activate",
+        &accel_group,
+        b'q' as u32,
```
EPashkin: IMHO better use http://gtk-rs.org/docs/gdk/enums/key/constant.q.html

src/bin/menu_bar.rs
FreeFull: It seems like `menu.set_accel_group` does two things: it makes the `accel_group` available through `menu.get_accel_group` and it refreshes the `accel_path`s for the menu and its children. Since we don't seem to need either, I decided to get rid of the line.

src/bin/menu_bar.rs
FreeFull: Thanks for the suggestion. I ended up using `gtk::accelerator_parse` to get access to the `<Primary>` modifier.

src/bin/menu_bar.rs
FreeFull: I have changed the code so it does actually use `Ctrl+Q` on Windows and Linux, and `Cmd+Q` on macOS. Helpfully, when you actually run the example, the key combo required is displayed alongside the menu item's name by GTK .

GuillaumeGomez:
Looks good to me, thanks!

Waiting for confirmation from @EPashkin and @sdroege and then we're good to go!

EPashkin:
@FreeFull Thanks
👍

sdroege:
If it works I guess it's fine, I don't know at all how this API is supposed to be used. If in doubt, ask some GTK people on IRC :)

PR #180, opened 2018-07-25
Improve code, remove clone when unnecessary

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/src/bin/grid.rs
+++ b/src/bin/grid.rs
@@ -43,22 +43,22 @@ mod example {
         window.set_application(application);
         let grid: Grid = builder.get_object("grid").expect("Couldn't get grid");
         let button6: Button = builder.get_object("button6").expect("Couldn't get button6");
-        button6.connect_clicked(clone!(grid, button6 => move |_| {
-            let height = grid.get_cell_height(&button6);
+        button6.connect_clicked(clone!(grid => move |button| {
+            let height = grid.get_cell_height(button);
             let new_height = if height == 2 { 1 } else { 2 };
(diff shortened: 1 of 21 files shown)


src/bin/grid.rs
```diff
@@ -43,22 +43,22 @@ mod example {
         window.set_application(application);
         let grid: Grid = builder.get_object("grid").expect("Couldn't get grid");
         let button6: Button = builder.get_object("button6").expect("Couldn't get button6");
-        button6.connect_clicked(clone!(grid, button6 => move |_| {
-            let height = grid.get_cell_height(&button6);
+        button6.connect_clicked(clone!(grid => move |button| {
```
sdroege: Is the button inside the grid? If so then this still leads to a reference cycle (use a weak ref for the grid)

sdroege:
Looks good generally

EPashkin:
@GuillaumeGomez Thanks

src/bin/grid.rs
GuillaumeGomez: What do you mean by weak ref? The `WeakRef` type or something else?

sdroege: Yes, call `downgrade()` before, pass the `WeakRef` into the closure, call `upgrade()` inside the closure and if it returns `None` simply do nothing

sdroege:
Looks good to me. I'll check later if I find other reference cycles in that code but merge for now :)

GuillaumeGomez:
Thanks for reviews!

Issue #179, opened 2018-07-25
Check for and fix all reference cycles in closures


sdroege:
One very direct example is here: https://github.com/gtk-rs/examples/blob/30549d056f21eafba755f8084f3cb8a85e4f6736/src/bin/clock.rs#L46-L49

The window will never be destroyed here because it keeps another strong reference to itself via the closure. Here this is completely unneeded even, the first parameter to the closure would already be the window.

In other cases we should use weak references instead if the data is not available from somewhere else.

----

The same problem can also happen in other places with `Arc`/`Rc`, and also indirectly between `GObject`s and those types. We should check all closures if they directly or indirectly (A has a reference to B, B has a reference to C, C has a reference to A) creates a reference cycle.

Checking all calls to `clone()` and usages of the `clone!` macro would be a good start.

sdroege:
Most are fixed now but e.g. all usages of the `clone!` macro in gtktest are wrong still. Probably in some other examples too

GuillaumeGomez:
This is worth a new PR I guess! ;)

PR #182, opened 2018-07-30
Fix various reference cycles in the examples

sdroege:
Fixes https://github.com/gtk-rs/examples/issues/179

### DIFF
--- a/src/bin/grid.rs
+++ b/src/bin/grid.rs
@@ -15,6 +15,19 @@ mod example {
 
     use std::env::args;
 
+    // upgrade weak reference or return
+    #[macro_export]
+    macro_rules! upgrade_weak {
+        ($x:ident, $r:expr) => {{
+            match $x.upgrade() {
+                Some(o) => o,
+                None => return $r,
+            }
+        }};
+        ($x:ident) => {
+            upgrade_weak!($x, ())
+        };
+    }
 
     pub fn build_ui(application: &gtk::Application) {
         let glade_src = include_str!("grid.glade");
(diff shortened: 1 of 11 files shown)


src/bin/grid.rs
```diff
@@ -15,6 +15,19 @@ mod example {
 
     use std::env::args;
 
+    // upgrade weak reference or return
+    #[macro_export]
+    macro_rules! upgrade_weak {
```
GuillaumeGomez: 👍

GuillaumeGomez:
Thanks! At some point, we'll also need to explain where people should `clone` and where they should `downgrade`/`upgrade`.

src/bin/listbox_model.rs
```diff
@@ -140,7 +159,9 @@ fn build_ui(application: &gtk::Application) {
 
         // When a row is activated (select + enter) we simply emit the clicked
         // signal on the corresponding edit button to open the edit dialog
-        box_.connect_activate(clone!(edit_button => move |_| {
+        let edit_button_weak = edit_button.downgrade();
+        box_.connect_activate(clone!(edit_button_weak => move |_| {
```
EPashkin: There was really need use `clone!` ?

src/bin/grid.rs
EPashkin: Seems this enough specific macro to be moved to glib

src/bin/listbox_model.rs
sdroege: No, actually not

src/bin/grid.rs
sdroege: It also works on `Arc` and `Rc` btw

GuillaumeGomez: Indeed, I'll let one of you add it to glib. ;) (and update examples after hehe)

PR #183, opened 2018-07-31
Remove some unneeded clones

sdroege:


src/bin/listbox_model.rs
```diff
@@ -139,10 +139,10 @@ fn build_ui(application: &gtk::Application) {
             // Activating the entry (enter) will send response 0 to the dialog, which
             // is the response code we used for the Close button. It will close the dialog
             let dialog_weak = dialog.downgrade();
-            entry.connect_activate(clone!(dialog_weak => move |_| {
+            entry.connect_activate(move |_| {
```
EPashkin: Can we replace `let dialog_weak = dialog.downgrade(); entry.connect_activate(clone!(dialog_weak...`
with macro analog of `clone`? Maybe even with auto `upgrade_weak!` for closures, returned ()

EPashkin: Or even allow `clone` macro do both `.clone` and `.downgrade` (Not sure that it possible)

sdroege: Yes that should be possible but I don't have time for that right now :)

I was thinking of a macro that can do it all: `clone!(strong: local_name@self.foo, weak: local_name2@self.bar, weak: local_name3@meh, weak_return: SomeReturnValue => move || {})` (with strong being the default, i.e. can be omitted and `weak_return` optional and then mapping to `()`)

Not sure if we need something else for matching there, but that

sdroege:
@GuillaumeGomez good to go :)

GuillaumeGomez:
Thanks!

PR #186, opened 2018-08-30
Finish feature "futures-nightly" removal

EPashkin:
@GuillaumeGomez PR in high-probable case that https://github.com/gtk-rs/glib/pull/373 fails again

cc @sdroege


EPashkin:
Strange, but https://github.com/gtk-rs/glib/pull/373 builds fine without this.
Build failure is just network error.
IMHO better merge this anyway, but seems it also safe to just close this PR.
@sdroege, @GuillaumeGomez what you think?

sdroege:
I wanted to keep as much as possible in place so it can easily be added back later, but I don't really care much. Whatever seems better to you :)

GuillaumeGomez:
If it works, it works so sounds good to me.

EPashkin:
Ok, I close this, as it don't break build, `gio_futures_await` still needed "futures-nightly".

GuillaumeGomez:
I merge then.

PR #191, opened 2018-09-23
Add status icon example

### LINKED ISSUE #158: add a StatusIcon example

GuillaumeGomez:
Fixes #158.

cc @sdroege @EPashkin 

Apparently, status icons (also called tray icons) won't be in gtk 4 and GNOME doesn't want them anymore. Also, they work very badly on systems I tested them on (linux and OSX) but they work. Also, they don't seem to work with `Application`. We really have a relic in here haha.


src/bin/status_icon.rs
```diff
@@ -0,0 +1,89 @@
+//! # TreeView Sample
+//!
+//! This sample demonstrates how to create a StatusIcon.
+
+extern crate gdk_pixbuf;
```
EPashkin: IMHO this file needed `#![allow(deprecated)]`, see https://travis-ci.org/gtk-rs/examples/jobs/432129991#L721

EPashkin:
I confirm that it works on windows
Looks good, except instead dialog IMHO better be single window, but as it relic this do as is.

sdroege:
Why add an example for deprecated and soon to disappear API? :)

GuillaumeGomez:
To close an issue. Like that it's done.

Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "gtk-rs-examples"
 version = "0.0.1"
 authors = ["The Gtk-rs Project Developers"]
+autobins = false
```
sdroege: Why?

src/bin/status_icon.rs
```diff
@@ -0,0 +1,91 @@
+//! # TreeView Sample
```
sdroege: Copy&paste mistake

src/bin/status_icon.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    menu.append(&dialog_item);
+    menu.append(&close_item);
+
+    let is_shown = Rc::new(RefCell::new(false));
+
+    let c_is_shown = is_shown.clone();
+    dialog_item.connect_activate(move |_| {
+        *c_is_shown.borrow_mut() = false;
+        let parent: Option<&gtk::Window> = None; // stupid hack to avoid rustc error
```
sdroege: Or: `None::<&gtk::Window>`

Cargo.toml
EPashkin: IMHO to prevent attempt to build gio_futures_await.rs in future
```
warning: An explicit [[bin]] section is specified in Cargo.toml which currently
disables Cargo from automatically inferring other binary targets.
This inference behavior will change in the Rust 2018 edition and the following
files will be included as a binary target:

* D:\eap\rust\0\examples\src\bin\gio_futures_await.rs

This is likely to break cargo build or cargo test as these files may not be
ready to be compiled as a binary target today. You can future-proof yourself
and disable this warning by adding `autobins = false` to your [package]
section. You may also move the files to a location where Cargo would not
automatically infer them to be a target, such as in subfolders.

For more information on this warning you can consult
https://github.com/rust-lang/cargo/issues/5330

```

GuillaumeGomez: @EPashkin: Exactly! Well found. ;)

src/bin/status_icon.rs
GuillaumeGomez: Good catch!

src/bin/status_icon.rs
GuillaumeGomez: I \*think\* that I had an issue with this syntax. I'll recheck just to be sure.

EPashkin:
@GuillaumeGomez as you will do change this PR again, please add `, "gio_2_40"` to `#default` in Cargo.toml too.

GuillaumeGomez:
@EPashkin You're sure? I don't mind doing it, just want a last check.

EPashkin:
Yes

GuillaumeGomez:
Then I'll do it as well.

EPashkin:
Thanks in advance

src/bin/status_icon.rs
sdroege: Fewer letters if it works :)

sdroege:
> To close an issue. Like that it's done.

You could've also closed it with an explanation why :)

src/bin/status_icon.rs
GuillaumeGomez: We all agree on that haha.

GuillaumeGomez:
@sdroege Naaaah. I'm pretty sure I'll need something like that some day. At least now it's done. :D

GuillaumeGomez:
(Oh and updated btw!)

sdroege:
Then go ahead :)

PR #193, opened 2018-09-26
Use property bindings instead of signals in the sync_widgets example

sdroege:


EPashkin:
Cool, I don't know about this possibility.
👍

GuillaumeGomez:
I didn't know about it either. Thanks!

PR #195, opened 2018-10-01
Use renamed GtkMenuExt

EPashkin:
Use changes from https://github.com/gtk-rs/gtk/pull/712 https://github.com/gtk-rs/gtk/pull/714

cc @GuillaumeGomez


EPashkin:
@GuillaumeGomez added it first

PR #197, opened 2018-10-13
Update responsetype

GuillaumeGomez:
cc @EPashkin  @sdroege

### DIFF
--- a/src/bin/gtktest.rs
+++ b/src/bin/gtktest.rs
@@ -112,8 +112,12 @@ mod example {
             let window = upgrade_weak!(window_weak);
             let entry = upgrade_weak!(entry_weak);
 
-            let dialog = Dialog::new_with_buttons(Some("Hello!"), Some(&window), gtk::DialogFlags::MODAL,
-                &[("No", 0), ("Yes", 1), ("Yes!", 2)]);
+            let dialog = Dialog::new_with_buttons(Some("Hello!"),
+                                                  Some(&window),
+                                                  gtk::DialogFlags::MODAL,
+                                                  &[("No", ResponseType::No),
+                                                    ("Yes", ResponseType::Yes),
+                                                    ("Custom", ResponseType::Other(0))]);
 
             let ret = dialog.run();
 
--- a/src/bin/listbox_model.rs
+++ b/src/bin/listbox_model.rs
@@ -23,6 +23,8 @@ extern crate gobject_subclass;
 use gio::prelude::*;
 use gtk::prelude::*;
 
+use gtk::ResponseType;
+
 use std::env::args;
 
 use row_data::RowData;
@@ -121,8 +123,8 @@ fn build_ui(application: &gtk::Application) {
             let window = upgrade_weak!(window_weak);
 
             let dialog = gtk::Dialog::new_with_buttons(Some("Edit Item"), Some(&window), gtk::DialogFlags::MODAL,
-                &[("Close", 0)]);
-            dialog.set_default_response(0);
+                &[("Close", ResponseType::Close)]);
+            dialog.set_default_response(ResponseType::Close);
             dialog.connect_response(|dialog, _| dialog.destroy());
 
             let content_area = dialog.get_content_area();
@@ -141,7 +143,7 @@ fn build_ui(application: &gtk::Application) {
             let dialog_weak = dialog.downgrade();
             entry.connect_activate(move |_| {
                 let dialog = upgrade_weak!(dialog_weak);
-                dialog.response(0);
+                dialog.response(ResponseType::Close);
             });
             content_area.add(&entry);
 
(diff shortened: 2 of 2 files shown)


src/bin/gtktest.rs
```diff
@@ -112,8 +112,12 @@ mod example {
...
             let entry = upgrade_weak!(entry_weak);
 
-            let dialog = Dialog::new_with_buttons(Some("Hello!"), Some(&window), gtk::DialogFlags::MODAL,
-                &[("No", 0), ("Yes", 1), ("Yes!", 2)]);
+            let dialog = Dialog::new_with_buttons(Some("Hello!"),
+                                                  Some(&window),
+                                                  gtk::DialogFlags::MODAL,
+                                                  &[("No", ResponseType::Other(0)),
+                                                    ("Yes", ResponseType::Other(1)),
+                                                    ("Yes!", ResponseType::Other(2))]);
```
EPashkin: Maybe better declare global statics or constants?

EPashkin:
Thanks again, 👍 even without constants

src/bin/gtktest.rs
GuillaumeGomez: Didn't think about it but I could yes. It just seems a bit too much something used only once. What do you think @sdroege?

src/bin/listbox_model.rs
```diff
@@ -121,8 +123,8 @@ fn build_ui(application: &gtk::Application) {
             let window = upgrade_weak!(window_weak);
 
             let dialog = gtk::Dialog::new_with_buttons(Some("Edit Item"), Some(&window), gtk::DialogFlags::MODAL,
-                &[("Close", 0)]);
-            dialog.set_default_response(0);
+                &[("Close", ResponseType::Other(0))]);
+            dialog.set_default_response(ResponseType::Other(0));
```
sdroege: `ReponseType::Close` exists, so why not use that? :)

src/bin/listbox_model.rs
```diff
@@ -141,7 +143,7 @@ fn build_ui(application: &gtk::Application) {
             let dialog_weak = dialog.downgrade();
             entry.connect_activate(move |_| {
                 let dialog = upgrade_weak!(dialog_weak);
-                dialog.response(0);
+                dialog.response(ResponseType::Other(0));
```
sdroege: `ResponseType::Close`

src/bin/listbox_model.rs
```diff
@@ -185,16 +187,16 @@ fn build_ui(application: &gtk::Application) {
             let window = upgrade_weak!(window_weak);
 
             let dialog = gtk::Dialog::new_with_buttons(Some("Add Item"), Some(&window), gtk::DialogFlags::MODAL,
-                &[("Ok", 0), ("Cancel", 1)]);
-            dialog.set_default_response(0);
+                &[("Ok", ResponseType::Other(0)), ("Cancel", ResponseType::Other(1))]);
```
sdroege: `ResponseType::Ok`, `::Cancel`

src/bin/listbox_model.rs
```diff
@@ -185,16 +187,16 @@ fn build_ui(application: &gtk::Application) {
...
+            dialog.set_default_response(ResponseType::Other(0));
```
sdroege: `ResponseType::Ok`

src/bin/listbox_model.rs
```diff
@@ -185,16 +187,16 @@ fn build_ui(application: &gtk::Application) {
...
 
             let content_area = dialog.get_content_area();
 
             let entry = gtk::Entry::new();
             let dialog_weak = dialog.downgrade();
             entry.connect_activate(move |_| {
                 let dialog = upgrade_weak!(dialog_weak);
-                dialog.response(0);
+                dialog.response(ResponseType::Other(0));
```
sdroege: `ResponseType::Ok`

src/bin/listbox_model.rs
```diff
@@ -203,7 +205,7 @@ fn build_ui(application: &gtk::Application) {
 
             dialog.connect_response(clone!(model, entry, spin_button => move |dialog, resp| {
                 if let Some(text) = entry.get_text() {
-                    if !text.is_empty() && resp == 0 {
+                    if !text.is_empty() && resp == ResponseType::Other(0) {
```
sdroege: `ResponseType::Ok`

src/bin/gtktest.rs
sdroege: There are `ResponseType::Yes` and `::No`, why not use those? :) And for the third one, maybe just call it `"Custom"` and use `::Other(0)` or so for that?

GuillaumeGomez:
@sdroege It was so obvious... I feel dumb for not thinking about. Thanks a lot for catching it!

src/bin/listbox_model.rs
```diff
@@ -141,7 +143,7 @@ fn build_ui(application: &gtk::Application) {
             let dialog_weak = dialog.downgrade();
             entry.connect_activate(move |_| {
                 let dialog = upgrade_weak!(dialog_weak);
-                dialog.response(0);
+                dialog.response(ResponseType::Close);
```
EPashkin: Still 0 here "send response 0 to the dialog"

EPashkin: I meant comment above

PR #199, opened 2018-10-25
Change event for application to build its UI

EPashkin:
Fixes problem described in https://github.com/gtk-rs/gtk/pull/731/files#r228063962
Also see https://wiki.gnome.org/HowDoI/GtkApplication,
https://developer.gnome.org/gtk3/stable/gtk-getting-started.html
https://gitlab.gnome.org/GNOME/gtk/blob/master/examples/bp/bloatpad.c

cc @GuillaumeGomez , @sdroege

### DIFF
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -34,7 +34,7 @@ fn build_ui(application: &gtk::Application) {
 
 fn main() {
     let application = gtk::Application::new("com.github.gtk-rs.examples.basic",
-                                            gio::ApplicationFlags::empty())
+                                            Default::default())
                                        .expect("Initialization failed...");
 
     application.connect_activate(|app| {
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -20,11 +20,6 @@ fn build_ui(application: &gtk::Application) {
     window.set_position(gtk::WindowPosition::Center);
     window.set_default_size(350, 70);
 
-    window.connect_delete_event(|win, _| {
-        win.destroy();
-        Inhibit(false)
-    });
-
     let button = gtk::Button::new_with_label("Click me!");
 
     window.add(&button);
--- a/src/bin/basic.rs
+++ b/src/bin/basic.rs
@@ -20,7 +20,7 @@ fn build_ui(application: &gtk::Application) {
     window.set_position(gtk::WindowPosition::Center);
     window.set_default_size(350, 70);
 
-    window.connect_delete_event(move |win, _| {
+    window.connect_delete_event(|win, _| {
         win.destroy();
         Inhibit(false)
     });
@@ -33,14 +33,13 @@ fn build_ui(application: &gtk::Application) {
 }
 
 fn main() {
-    let application = gtk::Application::new("com.github.basic",
+    let application = gtk::Application::new("com.github.gtk-rs.examples.basic",
                                             gio::ApplicationFlags::empty())
                                        .expect("Initialization failed...");
 
-    application.connect_startup(|app| {
+    application.connect_activate(|app| {
         build_ui(app);
     });
-    application.connect_activate(|_| {});
 
     application.run(&args().collect::<Vec<_>>());
 }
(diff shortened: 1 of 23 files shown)


EPashkin:
Also I start to think that we need version of `gtk::Application::new` without `ApplicationFlags`,
or implement `Default::default` for `ApplicationFlags` and use it.

EPashkin:
See https://github.com/gtk-rs/gtk/pull/731#discussion_r228080362
Here also can be other update with removing `connect_delete_event`

src/bin/basic.rs
```diff
@@ -20,7 +20,7 @@ fn build_ui(application: &gtk::Application) {
     window.set_position(gtk::WindowPosition::Center);
     window.set_default_size(350, 70);
 
-    window.connect_delete_event(move |win, _| {
+    window.connect_delete_event(|win, _| {
         win.destroy();
```
sdroege: You probably want app.quit() here?

sdroege:
Default:: default() for the flags imho

src/bin/basic.rs
EPashkin: I want remove this call completely: on windows it seems works same.
When main window closed it will be destroyed and application quits.
Ex. no connecting to "delete" event in https://gitlab.gnome.org/GNOME/gtk/tree/master/examples
and `g_application_quit` called only from normal action (timeouts, clicks, network commands)

EPashkin: Hm, I checked wrong event: we use "delete-event", but it even don't found in gtk or glib sources while `connect_delete_event` still called.

EPashkin:
Bad that after implementing `Default::default` for `ApplicationFlags` including `gio::prelude' or `gio::ApplicationExtXXX` still needed.
This will be true if we implement `withDefaultFlags` too :(

src/bin/basic.rs
EPashkin: yes, I thought that it don't needed as not in demos\examples
but I also have only gtk4 locally too, so not sure now.

EPashkin:
Updated

src/bin/menu_bar_system.rs
```diff
@@ -9,7 +9,7 @@ extern crate gtk;
 use gio::prelude::*;
 use gtk::{
     AboutDialog, AboutDialogExt, BoxExt, ContainerExt, DialogExt, GtkApplicationExt, GtkWindowExt,
-    Inhibit, LabelExt, SwitchExt, ToVariant, WidgetExt,
+    LabelExt, SwitchExt, ToVariant, WidgetExt,
```
sdroege: Why don't you just import the gtk prelude here (and in other places)?

sdroege:
> Remove unneeded connect_delete_event's

Are there any needed ones left? Just curious. Looks good except for that prelude thing :)

EPashkin:
https://github.com/gtk-rs/examples/blob/master/src/bin/multi_windows.rs#L54-L57 still remains

src/bin/menu_bar_system.rs
EPashkin: Mostly because it was already imported this way :wink:
Want me change this too ?

sdroege: Yes please :)

sdroege:
All good for me then

EPashkin:
Added `use gtk::prelude::*;` instead import separate traits.
Leaved out `use gio::{ApplicationExt, ApplicationExtManual};` in menu_bar.rs as it only 2
and `use gdk::ScreenExt;`in transparent_main_window.rs.
Can change this too

sdroege:
> Leaved out `use gio::{ApplicationExt, ApplicationExtManual};` in menu_bar.rs as it only 2
> and `use gdk::ScreenExt;`in transparent_main_window.rs.
> Can change this too

Please do. It's not nice to require users to know about these ext traits :)

EPashkin:
Ok, changed this 2 too.

GuillaumeGomez:
Thanks!

PR #202, opened 2018-11-07
Add overlay example

### LINKED ISSUE #200: Add overlay example

GuillaumeGomez:
Fixes #200.

cc @sdroege @EPashkin

### DIFF
--- a/src/bin/overlay.rs
+++ b/src/bin/overlay.rs
@@ -1,0 +1,120 @@
+//! # Overlay example
+//!
+//! This sample demonstrates how to create an element "floating" above others.
+
+extern crate gdk;
+extern crate gio;
+extern crate glib;
+extern crate gtk;
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+use std::env::args;
+
+// Basic CSS: we change background color, we set font color to black and we set it as bold.
+const STYLE: &'static str = "
+#overlay-label {
+    background-color: rgba(192, 192, 192, 0.8);
+    color: black;
+    font-weight: bold;
+}";
+
+// upgrade weak reference or return
+#[macro_export]
+macro_rules! upgrade_weak {
+    ($x:ident, $r:expr) => {{
+        match $x.upgrade() {
+            Some(o) => o,
+            None => return $r,
+        }
+    }};
+    ($x:ident) => {
+        upgrade_weak!($x, ())
+    };
+}
+
+fn button_clicked(button: &gtk::Button, overlay_text_weak: &glib::object::WeakRef<gtk::Label>) {
+    let overlay_text = upgrade_weak!(overlay_text_weak);
+    overlay_text.set_text(&button.get_label().expect("Couldn't get button label"));
+}
+
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title("Overlay");
+    window.set_position(gtk::WindowPosition::Center);
+
+    // The overlay container.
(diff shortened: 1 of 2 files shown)


src/bin/overlay.rs
```diff
@@ -0,0 +1,123 @@
...
+
+use std::env::args;
+
+// Basic CSS: we change background color, we set font color to black and we set it as bold.
+const STYLE: &'static str = "
+#countdown-label {
+    background-color: rgba(192, 192, 192, 0.8);
+    color: black;
+    font-weight: bold;
+}";
```
sdroege: I know this CSS ;)

src/bin/overlay.rs
```diff
@@ -0,0 +1,123 @@
...
+        Inhibit(false)
+    });
+
+    // The overlay container.
+    let overlay = gtk::Overlay::new();
+
+    // The overlay label.
+    let overlay_text = gtk::Label::new("0");
+    // We need to name it in order to apply CSS on it.
+    gtk::WidgetExt::set_name(&overlay_text, "countdown-label");
```
sdroege: Why is it called countdown if we don't do one?

src/bin/overlay.rs
```diff
@@ -0,0 +1,123 @@
...
+    overlay_text.set_halign(gtk::Align::End);
+    overlay_text.set_valign(gtk::Align::Start);
+
```
sdroege: Why not centered?

src/bin/overlay.rs
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }};
+    ($x:ident) => {
+        upgrade_weak!($x, ())
+    };
+}
+
+fn button_clicked(button: &gtk::Button, overlay_text_weak: &glib::object::WeakRef<gtk::Label>) {
+    let overlay_text = upgrade_weak!(overlay_text_weak);
+    overlay_text.set_text(&button.get_label().unwrap_or_else(|| String::new()));
```
sdroege: The unwrap can never fail, so expect :)

src/bin/overlay.rs
EPashkin: IMHO current `set_halign(gtk::Align::End);` is better that centered.

EPashkin:
@GuillaumeGomez Thanks

src/bin/overlay.rs
GuillaumeGomez: Where could it come from?! :D

src/bin/overlay.rs
GuillaumeGomez: Ah true. I'll rename it.

src/bin/overlay.rs
GuillaumeGomez: Sounds better for the current case.

src/bin/overlay.rs
GuillaumeGomez: Yes sir!

src/bin/overlay.rs
```diff
@@ -0,0 +1,123 @@
...
+            .expect("Failed to load CSS");
+        gtk::StyleContext::add_provider_for_screen(
+            &gdk::Screen::get_default().expect("Error initializing gtk css provider."),
+            &provider,
+            gtk::STYLE_PROVIDER_PRIORITY_APPLICATION,
+        );
+
+        // We build the application UI.
+        build_ui(app);
+    });
```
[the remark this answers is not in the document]
GuillaumeGomez: Ok!

sdroege:
:+1:

EPashkin:
👍

PR #204, opened 2018-11-15
Add atk/accessibility example

GuillaumeGomez:
cc @EPashkin @sdroege


EPashkin:
I don't understand how it works on Windows,
but at minimum code runs without panic.
@GuillaumeGomez, Thanks.

PR #216, opened 2019-01-10
Add progress-tracker example

### LINKED ISSUE #209: Progress bar example when button is pressed
when I try to do progress bar update when a button is clicked, nothing is updated until the button click code is completed. Does any one have an example for using the Progress Bar?

mmstick:
Adds example with a progress bar that's updated over time after clicking on a button. The button spawns a thread that sends an event each second through a channel to a `gtk::idle_add` that periodically checks for values.

Adds the cascade crate since it makes constructing GTK widgets simpler.

Closes #209

### DIFF
--- a/src/bin/progress_tracker.rs
+++ b/src/bin/progress_tracker.rs
@@ -1,0 +1,211 @@
+//! Track progress with a background thread and a channel.
+
+extern crate gio;
+extern crate glib;
+extern crate gtk;
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+use std::cell::Cell;
+use std::env::args;
+use std::rc::Rc;
+use std::sync::mpsc::{self, TryRecvError};
+use std::thread;
+use std::time::Duration;
+
+pub fn main() {
+    glib::set_program_name("Progress Tracker".into());
+
+    let application = gtk::Application::new(
+        "com.github.progress-tracker",
+        gio::ApplicationFlags::empty(),
+    ).expect("initialization failed");
+
+    application.connect_startup(|app| {
+        Application::new(app);
+    });
+
+    application.connect_activate(|_| {});
+    application.run(&args().collect::<Vec<_>>());
+}
+
+pub struct Application {
+    pub widgets: Rc<Widgets>,
+}
+
+impl Application {
+    pub fn new(app: &gtk::Application) -> Self {
+        let app = Application {
+            widgets: Rc::new(Widgets::new(app)),
+        };
+
+        app.connect_progress();
+
+        app
+    }
+
+    fn connect_progress(&self) {
+        let widgets = self.widgets.clone();
+        let active = Rc::new(Cell::new(false));
+        self.widgets.main_view.button.connect_clicked(move |_| {
+            let active = active.clone();
+            if active.get() {
+                return;
+            }
+
+            active.set(true);
+
+            let (tx, rx) = mpsc::channel();
+            thread::spawn(move || {
+                for v in 1..=10 {
+                    let _ = tx.send(v);
+                    thread::sleep(Duration::from_millis(500));
+                }
+            });
+
+            let widgets = widgets.clone();
+            gtk::timeout_add(16, move || match rx.try_recv() {
+                Ok(value) => {
+                    widgets
+                        .main_view
+                        .progress
+                        .set_fraction(f64::from(value) / 10.0);
+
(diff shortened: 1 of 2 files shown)


src/bin/progress_tracker.rs
```diff
@@ -0,0 +1,208 @@
...
+    pub main_view: MainView,
+    pub complete_view: CompleteView,
+}
+
+impl Widgets {
+    pub fn new(application: &gtk::Application) -> Self {
+        let complete_view = CompleteView::new();
+        let main_view = MainView::new();
+
+        let view_stack = cascade! {
```
GuillaumeGomez: I'm not a big fan of using a macro to do something so basic. Repeating `view_stack` everytime is annoying, I agree. However, the goal here is to provide some code as easy to understand as possible. The `cascade` macro clearly doesn't help on this point.

GuillaumeGomez: Even worse: it actually doesn't make the code shorter (just realized it afterwards).

mmstick: > Repeating view_stack everytime is annoying, I agree. However, the goal here is to provide some code as easy to understand as possible. The cascade macro clearly doesn't help on this point.

I disagree. The cascade pattern is easy to understand, as it's no different than the builder pattern. The point of the builder and cascade pattern is to prevent repeating yourself. It also encourages grouping declarations together into the same scope. Without the cascade pattern, the GTK API is awkward / unergonomic to use.

> Even worse: it actually doesn't make the code shorter (just realized it afterwards).

If by shorter you mean it uses an additional line or two, sure, since Rust doesn't yet support the cascade pattern. If by shorter you mean it requires less typing and reading, then no, the cascade pattern is significantly shorter.

The pattern is used in a few other popular programming languages, such as Google's Dart, so I don't think it's confusing to run across, especially since we already use the builder pattern, which is virtually identical. GTK doesn't have a builder pattern, so the cascade pattern works just as well.

GuillaumeGomez: > I disagree. The cascade pattern is easy to understand, as it's no different than the builder pattern. The point of the builder and cascade pattern is to prevent repeating yourself. It also encourages grouping declarations together into the same scope. Without the cascade pattern, the GTK API is awkward / unergonomic to use.

Maybe but it's not the standard documentation. Readers are already trying to learn how to use gtk-rs, let's not improve the difficulty.

mmstick: > Maybe but it's not the standard documentation.

You can add it to the standard documentation.

> Readers are already trying to learn how to use gtk-rs, let's not improve the difficulty.

I'd argue that it is decreasing the difficulty, not making it more difficult. Learning how to use cascade is absolutely miniscule in comparison to gtk-rs. It's hardly different from using the builder pattern in existing Rust std APIs. Once learned, it makes working with gtk-rs significantly easier.

mmstick: Documentation and references can be added anywhere. Whether it is documented in this repo or elsewhere.

> If users want to use more advanced rust's features, it's up to them.

Many do not realize that they can use the method-cascade pattern in Rust, or may not yet be familiar with the existence of the method-cascades. I'd rather see the proper patterns being used and learned, rather than limiting these examples to using the least-idiomatic syntax.

Anyway, I strongly believe that cascade should be used, so I will close this PR until Rust gains official support for this feature.

GuillaumeGomez: That's a shame since your example is great. :-/

Could you reopen a PR without the cascade feature for now and once rust has it natively, update it? That sounds like a good compromise for me.

mmstick: I guess so. When I have free time, I'l see about rewriting the cascade parts.

GuillaumeGomez: Thanks a lot!

EPashkin:
@mmstick Thanks, but there some nagging:
"gtk/v3_22" is IMHO too hight, "gtk_3_10" is seems enough (https://gtk-rs.org/docs/gtk/struct.HeaderBar.html)
better change name to match source file,
if it not problem, can you fix git tree too, now new file added in merge commit that very strange and hard to find later.

GuillaumeGomez:
Is it ready to go?

mmstick:
As far as I know.

GuillaumeGomez:
Let's merge then. Thanks!

PR #220, opened 2019-01-24
Add a drag-and-drop example

osa1:
Ping @sdroege

### DIFF
--- a/src/bin/drag_and_drop_listbox.rs
+++ b/src/bin/drag_and_drop_listbox.rs
@@ -1,0 +1,189 @@
+extern crate cairo;
+extern crate gdk;
+extern crate gio;
+extern crate gtk;
+
+use gdk::{Atom, DragAction, DragContext, ModifierType, Screen};
+use gio::prelude::*;
+use gtk::*;
+use std::env::args;
+
+fn main() {
+    let app = gtk::Application::new("com.github.gtk-rs.examples.basic", Default::default())
+        .expect("Initialization failed...");
+
+    app.connect_activate(|app| {
+        build_ui(app);
+    });
+
+    app.run(&args().collect::<Vec<_>>());
+}
+
+fn build_ui(app: &gtk::Application) {
+    let css_provider = CssProvider::new();
+    css_provider.load_from_data(CSS.as_bytes()).unwrap();
+    StyleContext::add_provider_for_screen(
+        Screen::get_default().as_ref().unwrap(),
+        &css_provider,
+        STYLE_PROVIDER_PRIORITY_APPLICATION,
+    );
+
+    let window = gtk::ApplicationWindow::new(app);
+    window.set_title("A drag-and-drop example");
+    window.set_default_size(350, 500);
+
+    let list_box = ListBox::new();
+    list_box.set_selection_mode(SelectionMode::None);
+
+    window.add(&list_box);
+
+    // Add rows
+    for i in 0..20 {
+        list_box.add(&create_row(i));
+    }
+
+    window.show_all();
+}
+
+fn create_row(i: i32) -> ListBoxRow {
+    let list_box_row = ListBoxRow::new();
+    let box_ = Box::new(Orientation::Horizontal, 5);
+    box_.set_margin_start(10);
+    box_.set_margin_end(10);
+    list_box_row.add(&box_);
+
+    let event_box = EventBox::new();
+    let image = Image::new_from_icon_name("open-menu-symbolic", IconSize::Button);
+    event_box.add(&image);
+    box_.add(&event_box);
+
+    let label_text = format!("Row {}", i);
+    let label = Label::new(Some(label_text.as_str()));
+
+    box_.add(&label);
+
+    // Drag stuff
+    //
+    // Make the event box draggable. Triggered with left mouse button.
+    //
+    let drag_targets: [TargetEntry; 1] = [TargetEntry::new(
+        "GTK_LIST_BOX_ROW", // I don't know how this string is used
+        TargetFlags::SAME_APP,
+        0,
+    )];
+    event_box.drag_source_set(ModifierType::BUTTON1_MASK, &drag_targets, DragAction::MOVE);
+    // drag_begin will be called when we start dragging. drag_begin is where we call rendering the
+    // widget being dragged next to mouse cursor.
+    event_box.connect_drag_begin(drag_begin);
+    // We don't use this, but just to demonstrate that things can be done when the drag ends.
+    event_box.connect_drag_end(drag_end);
+    // See https://developer.gnome.org/gtkmm-tutorial/stable/sec-dnd-signals.html.en
+    // Called when a source is being dragged and the target calls get_data()
+    event_box.connect_drag_data_get(drag_data_get);
+    // Set the same box as a drag destination.
+    event_box.drag_dest_set(DestDefaults::ALL, &drag_targets, DragAction::MOVE);
+    // Called when get_data() returns
+    event_box.connect_drag_data_received(drag_data_received);
+
+    list_box_row
+}
+
+fn drag_begin<W: IsA<Widget>>(w: &W, ctx: &DragContext) {
+    println!("drag_begin({:?})", w);
+
+    // Get containing ListBoxRow and render it next to the cursor while being dragged
+    match w.get_ancestor(ListBoxRow::static_type()) {
+        None => {
+            println!("Can't get ancestor");
+        }
+        Some(ancestor) => {
+            let alloc = ancestor.get_allocation();
+            let surface: cairo::ImageSurface =
+                cairo::ImageSurface::create(cairo::Format::ARgb32, alloc.width, alloc.height)
+                    .unwrap();
+            let surface_: &cairo::Surface = surface.as_ref();
+
+            {
+                ancestor.get_style_context().add_class("drag-icon");
+                let surface_ctx = cairo::Context::new(surface_);
+                ancestor.draw(&surface_ctx);
+                ancestor.get_style_context().remove_class("drag-icon");
+            }
+
+            let (x, y) = w.translate_coordinates(&ancestor, 0, 0).unwrap();
+            surface.set_device_offset(-x as f64, -y as f64);
+            ctx.drag_set_icon_surface(surface_);
+        }
+    }
+}
+
+fn drag_end<W: IsA<Widget>>(w: &W, _ctx: &DragContext) {
+    println!("drag_end({:?})", w);
+}
+
+fn drag_data_get<W: IsA<Widget>>(w: &W, _ctx: &DragContext, sel: &SelectionData, _: u32, _: u32) {
+    println!("drag_data_get({:?}", w);
+
+    // Get containing ListBoxRow index and put that into SelectionData, to be read in
+    // drag_data_received below
+    let list_box_row = w.get_ancestor(ListBoxRow::static_type()).unwrap();
+    // Downcast it
+    let list_box_row = list_box_row.downcast_ref::<ListBoxRow>().unwrap();
+
+    let row_idx = list_box_row.get_index();
+    sel.set(&Atom::intern(""), 32, &row_idx.to_ne_bytes());
+}
+
+fn drag_data_received<W: IsA<Widget>>(
+    target: &W,
+    _ctx: &DragContext,
+    _x: i32,
+    _y: i32,
+    sel: &SelectionData,
+    _info: u32,
+    _time: u32,
+) {
+    println!("drag_data_received({:?})", target);
+
+    // Index of the widget being dragged, as we stored in the SelectionData in drag_data_get()
+    // above.
+    let row_idx: i32 = {
+        let bytes = sel.get_data();
+        assert!(bytes.len() == 4);
+        let bytes: [u8; 4] = [bytes[0], bytes[1], bytes[2], bytes[3]];
+        i32::from_ne_bytes(bytes)
+    };
+
+    // Get the list box that contains both the source and the target
+    let list_box = target.get_ancestor(ListBox::static_type()).unwrap();
+    // Downcast it
+    let list_box: &ListBox = list_box.downcast_ref::<ListBox>().unwrap();
+
+    // Get the source widget from the list box
+    let source = list_box.get_row_at_index(row_idx).unwrap();
+    // Downcast it
+    let source = source.downcast_ref::<ListBoxRow>().unwrap();
+
+    // The target is a EventBox, get its containing ListBoxRow
+    let target = target.get_ancestor(ListBoxRow::static_type()).unwrap();
+    // Downcast it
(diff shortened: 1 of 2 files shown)


src/bin/listbox-dnd.rs
```diff
@@ -0,0 +1,228 @@
...
+    _ctx: &DragContext,
+    sel: &SelectionData,
+    _: u32,
+    _: u32,
+) {
+    println!("drag_data_get({:?}", w);
+
+    *widget_being_dragged = Some(w.as_ref().clone());
+    // We have to call this to emit drag_data_received() on the target
+    sel.set(&Atom::intern(""), 32, &[]);
```
sdroege: I believe you can simply store the index of the widget inside the listbox here and then later use `get_row_at_index()` to get the widget back from the listbox

src/bin/listbox-dnd.rs
```diff
@@ -0,0 +1,228 @@
...
+    if source == target {
+        println!("Source and target are the same. Aborting drag.");
+        return;
+    }
+
+    // Remove source from its parent
+    source
+        .get_parent()
+        .unwrap()
+        .downcast_ref::<Container>()
```
sdroege: The parent for source and target is the listbox or not? Why do you get it complicated like this instead of just directly accessing the listbox?

osa1: Widget being dragged and target are `EventBox`, not `ListBox`.

EPashkin:
Maybe rename it to "drag_and_drop_listbox.rs" match other dnd examples?

osa1:
Huh, I wasn't even aware that there are other dnd examples. I'll update the file name.

EPashkin:
It actually required gtk 3.12 due [set_margin_end](https://gtk-rs.org/docs/gtk/trait.WidgetExt.html#tymethod.set_margin_end) but you need define `gtk_3_12` before.
Also better use `gtk::Application::new` as direct `Window::new` approach is deprecated

EPashkin:
Window better changed from "First GTK+ Program" 😉

src/bin/listbox-dnd.rs
```diff
@@ -0,0 +1,228 @@
...
+        .get_parent()
+        .unwrap()
+        .downcast_ref::<ListBox>()
+        .unwrap()
+        .insert(source, target_row);
+}
+
+static CSS: &'static str = "
+    .drag-icon {
+        background: white;
```
EPashkin: How this works with dark theme?
Maybe there some alias used for color for listbox background color?

osa1:
> Also better use gtk::Application::new as direct Window::new approach is deprecated

Done

> Window title better changed from "First GTK+ Program"

Done

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,199 @@
...
+    window.set_default_size(350, 500);
+
+    let list_box = ListBox::new();
+    list_box.set_selection_mode(SelectionMode::None);
+
+    window.add(&list_box);
+
+    window.connect_delete_event(|_, _| {
+        gtk::main_quit();
+        Inhibit(false)
```
EPashkin: This unneeded with `ApplicationWindow`

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,199 @@
...
+    });
+
+    // Add rows
+    for i in 0..20 {
+        list_box.add(&create_row(i));
+    }
+
+    window.show_all();
+    gtk::main();
```
EPashkin: This too

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+        &css_provider,
+        STYLE_PROVIDER_PRIORITY_APPLICATION,
+    );
+
+    let window = gtk::ApplicationWindow::new(app);
+    window.set_title("A drag-and-drop example");
+    window.set_default_size(350, 500);
+
+    let list_box = ListBox::new();
+    list_box.set_selection_mode(SelectionMode::None);
```
sdroege: Shouldn't this be `::Single` so that a single item can be selected (and dragged)?

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+// gtk-rs by Ömer Sinan Ağacan <omeragacan@gmail.com>.
+
+extern crate cairo;
+extern crate gdk;
+extern crate gio;
+extern crate gtk;
+
+use gdk::{Atom, DragAction, DragContext, ModifierType, Screen};
+use gio::prelude::*;
+use gtk::*;
```
sdroege: Don't do wildcard imports from `gtk`, if only from the `prelude`. Especially for an example it's otherwise not really clear where types come from :)

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    box_.add(&label);
+
+    // Drag stuff
+    //
+    // Make the event box draggable. Triggered with left mouse button.
+    //
+    let drag_targets: [TargetEntry; 1] = [TargetEntry::new(
+        "GTK_LIST_BOX_ROW", // I don't know how this string is used
+        TargetFlags::SAME_APP,
```
sdroege: This one here is btw the reason why putting pointers into the selection data is a very bad idea (even if you're careful to get the memory management right). Things could be shared between different processes.

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+    let label = Label::new(Some(label_text.as_str()));
+
+    box_.add(&label);
+
+    // Drag stuff
+    //
+    // Make the event box draggable. Triggered with left mouse button.
+    //
+    let drag_targets: [TargetEntry; 1] = [TargetEntry::new(
+        "GTK_LIST_BOX_ROW", // I don't know how this string is used
```
sdroege: AFAIU this is the "type" of the data that can be provided as a drag target. I.e. here you say that you can accept things that have this specific type, and somewhere there's going to be some filtering to make sure you don't accept anything you can't handle

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+    list_box_row
+}
+
+fn drag_begin<W: IsA<Widget>>(w: &W, ctx: &DragContext) {
+    println!("drag_begin({:?})", w);
+
+    // Get containing ListBoxRow and render it next to the cursor while being dragged
+    match w.get_ancestor(ListBoxRow::static_type()) {
+        None => {
+            println!("Can't get ancestor");
```
sdroege: This can't possibly happen: `expect()` on it and get rid of a whole level of indentation :)

sdroege: Two levels of indentation actually

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+        }
+        Some(ancestor) => {
+            let alloc = ancestor.get_allocation();
```
sdroege: This code here and below probably needs some comments to explain what exactly is happening here

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+            let surface: cairo::ImageSurface =
+                cairo::ImageSurface::create(cairo::Format::ARgb32, alloc.width, alloc.height)
+                    .unwrap();
+            let surface_: &cairo::Surface = surface.as_ref();
```
sdroege: This should be unneeded. `ImageSurface` implements `Deref<Target=Surface>`, so you can directly call `Surface` methods on it

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+
+            {
+                ancestor.get_style_context().add_class("drag-icon");
+                let surface_ctx = cairo::Context::new(surface_);
+                ancestor.draw(&surface_ctx);
+                ancestor.get_style_context().remove_class("drag-icon");
+            }
+
+            let (x, y) = w.translate_coordinates(&ancestor, 0, 0).unwrap();
```
sdroege: Personally I'm fine with all this `unwrap()`s but @GuillaumeGomez will probably want them all changed to `expect()`s :)

GuillaumeGomez: Yes I do!

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+            surface.set_device_offset(-x as f64, -y as f64);
+            ctx.drag_set_icon_surface(surface_);
+        }
+    }
+}
+
+fn drag_end<W: IsA<Widget>>(w: &W, _ctx: &DragContext) {
+    println!("drag_end({:?})", w);
```
sdroege: Does the `Debug` impl of the `DragContext` print anything meaningful apart from the pointer? Maybe you want to print some other meaningful values here then :)

src/bin/drag_and_drop_listbox.rs
sdroege: Should probably figure this out before merging, can you ask in `#gtk+` on the GNOME IRC?

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+    println!("drag_data_get({:?}", w);
+
+    // Get containing ListBoxRow index and put that into SelectionData, to be read in
+    // drag_data_received below
+    let list_box_row = w.get_ancestor(ListBoxRow::static_type()).unwrap();
+    // Downcast it
+    let list_box_row = list_box_row.downcast_ref::<ListBoxRow>().unwrap();
+
+    let row_idx = list_box_row.get_index();
+    sel.set(&Atom::intern(""), 32, &row_idx.to_ne_bytes());
```
sdroege: Shouldn't we use a meaningful `Atom` here?

sdroege: Also instead of `32`, maybe `mem::size_of_val(row_idx) * 8`?

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+    sel: &SelectionData,
+    _info: u32,
+    _time: u32,
+) {
+    println!("drag_data_received({:?})", target);
+
+    // Index of the widget being dragged, as we stored in the SelectionData in drag_data_get()
+    // above.
+    let row_idx: i32 = {
+        let bytes = sel.get_data();
```
sdroege: You probably want to check the `Atom` and `format` fields here to make sure it's the data you expect

sdroege:
Thanks, generally looks quite nice :)

src/bin/drag_and_drop_listbox.rs
osa1: That might be better from UX point of view, but the original example that I ported did exactly this.

sdroege: Fine with me :) Does it change anything with the behaviour of the application?

src/bin/drag_and_drop_listbox.rs
```diff
@@ -0,0 +1,193 @@
...
+use gio::prelude::*;
+use gtk::*;
+use std::env::args;
+
+fn main() {
+    let app = gtk::Application::new("com.github.gtk-rs.examples.basic", Default::default())
+        .expect("Initialization failed...");
+
+    app.connect_activate(|app| {
+        build_ui(app);
```
sdroege: You want to use `startup` here instead of `activate`, see comment on the other PR

GuillaumeGomez:
@osa1 Still working on this? (We're getting close to the release and I'd like to merge the last things before releasing)

osa1:
@GuillaumeGomez I'll try to address comments later today.

src/bin/drag_and_drop_listbox.rs
osa1: This line prints `drag_end(EventBox { inner: ObjectRef(Shared { inner: 0x55b9f31ade20, borrowed: true }), type: GtkEventBox })`.

I used prints for debugging, I'll just remove them now.

src/bin/drag_and_drop_listbox.rs
osa1: > Does it change anything with the behaviour of the application?

No functional changes but it looks strange because you can't do anything with the selected items.

osa1:
Addressed most of the comments. I don't understand the remaining stuff so I'll need to do some reading for those.

src/bin/drag_and_drop_listbox.rs
osa1: I'll try to figure this out.

src/bin/drag_and_drop_listbox.rs
sdroege: If you call `get_format()` and `get_target()` on it first, you can compare that to the things you set in `drag_data_get()`. If the values are different then you got data from a different drag event

src/bin/drag_and_drop_listbox.rs
sdroege: Ack, let's just keep it as is then :)

sdroege:
Looks good to me now apart from

* https://github.com/gtk-rs/examples/pull/220#discussion_r250781061
* https://github.com/gtk-rs/examples/pull/220#discussion_r250777644
* https://github.com/gtk-rs/examples/pull/220#discussion_r250606064

For the last one about the CSS, there are things like e.g. `@selected_bg_color` or `@selected_fg_color` which are the colors defined by themes. Might be worth going through the CSS of themes to see what's there.

GuillaumeGomez:
There is a conflict. Beyond that, anything else that remains to updated?

sdroege:
> Looks good to me now apart from
> 
>     * [#220 (comment)](https://github.com/gtk-rs/examples/pull/220#discussion_r250781061)
> 
>     * [#220 (comment)](https://github.com/gtk-rs/examples/pull/220#discussion_r250777644)
> 
>     * [#220 (comment)](https://github.com/gtk-rs/examples/pull/220#discussion_r250606064)
> 
> 
> For the last one about the CSS, there are things like e.g. `@selected_bg_color` or `@selected_fg_color` which are the colors defined by themes. Might be worth going through the CSS of themes to see what's there.

None of these seem to be handled yet.

jsparber:
This isn't building for me. I guess because the gtk::Application::new api changed

sdroege:
This one needs to be updated for the new API and there are some pending comments (https://github.com/gtk-rs/examples/pull/220#issuecomment-462113988) that need checking with GTK folks for how to do it properly.

@osa1 are you planning to update this and do that?

osa1:
> @osa1 are you planning to update this and do that?

Nope. This can be closed.

PR #221, opened 2019-01-25
Add a list store example

osa1:


src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
+// "List Store" gtk3 demo ported to gtk-rs by Ömer Sinan Ağacan <omeragacan@gmail.com>
```
GuillaumeGomez: Please use same header as everywhere else (if we have one?). Git blame is there to know who wrote the code. ;)

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+
+use std::env::args;
+use std::rc::Rc;
+
+fn main() {
+    let application =
+        gtk::Application::new("com.github.gtk-rs.examples.list-store", Default::default())
+            .expect("Initialization failed...");
+
+    application.connect_activate(|app| {
```
GuillaumeGomez: The correct code is:

```rust
    application.connect_startup(|app| {
        build_ui(app);
    });
application.connect_activate(|_| {});
```

`connect_activate` is fow when the same binary is launched again so generally, you just want to show the window, not create a new one.

src/bin/list_store.rs
osa1: This looks fine to me. I don't think we have a standard header for examples. Lots of examples don't have any headers.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+        build_ui(app);
+    });
+
+    application.run(&args().collect::<Vec<_>>());
+}
+
+static COLUMN_FIXED: i32 = 0;
```
GuillaumeGomez: Rust has enums you know? :)

osa1: I don't need enums though, I need `i32`s. Can I convert a Rust enum variant to `i32`?

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    window.set_title("List Store");
+    window.set_border_width(10);
+    window.set_position(gtk::WindowPosition::Center);
+    window.set_default_size(280, 250);
+
+    let vbox = Box::new(Orientation::Vertical, 8);
+    window.add(&vbox);
+
+    let label = Label::new(Some(
```
GuillaumeGomez: No need for `Some`.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+    treeview.set_vexpand(true);
+    treeview.set_search_column(COLUMN_DESCRIPTION);
+
+    sw.add(&treeview);
+
+    add_columns(model.clone(), &treeview);
+
+    window.show_all();
+
+    {
```
GuillaumeGomez: Why creating a block here?

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+        ];
+        store.set(&store.append(), &col_indices, &values);
+    }
+
+    store
+}
+
+fn fixed_toggled<W: IsA<CellRendererToggle>>(model: &gtk::ListStore, _w: &W, path: TreePath) {
+    let iter = model.get_iter(&path).unwrap();
+    let mut fixed = model.get_value(&iter, COLUMN_FIXED).get::<bool>().unwrap();
```
GuillaumeGomez: `if let Some(x)` pattern seems to be a better idea in here.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+        column.add_attribute(&renderer, "sensitive", COLUMN_SENSITIVE);
+        column.set_sort_column_id(COLUMN_ICON);
+        treeview.append_column(&column);
+    }
+}
+
+fn spinner_timeout(model: &gtk::ListStore) -> Continue {
+    let iter = model.get_iter_first().unwrap();
+    let mut pulse = model.get_value(&iter, COLUMN_PULSE).get::<u32>().unwrap();
+    if pulse == u32::max_value() {
```
GuillaumeGomez: I recommend [`checked_add`](https://doc.rust-lang.org/nightly/std/primitive.u32.html#method.checked_add) instead.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+        pulse = 0;
+    } else {
+        pulse += 1;
+    }
+
+    model.set_value(&iter, COLUMN_PULSE as u32, &pulse.to_value());
```
GuillaumeGomez: Seems weird to create a const with a given type and then cast it...

GuillaumeGomez:
I have a personnal preference to put the `main` function at the end of the file (like in other examples). Also, please add more comments to explain what's going on in the code. Try to put yourself at the place of new users who don't know very well gtk-rs (or even gtk!).

src/bin/list_store.rs
GuillaumeGomez: Yep: `#[repr(i32)]`.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
+// "List Store" gtk3 demo ported to gtk-rs by Ömer Sinan Ağacan <omeragacan@gmail.com>
+
+extern crate gio;
+extern crate gtk;
+
+use gio::*;
+use gtk::*;
```
sdroege: Don't do wildcard imports for the whole crates, see my comment in the other example

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+    vbox.add(&sw);
+
+    let model = Rc::new(create_model());
+    let treeview = TreeView::new_with_model(&*model);
+    treeview.set_vexpand(true);
+    treeview.set_search_column(COLUMN_DESCRIPTION);
+
+    sw.add(&treeview);
+
+    add_columns(model.clone(), &treeview);
```
sdroege: Might be nicer to pass a `&Rc<_>` here and let the code in there worry about cloning or not if needed, but feel free to keep it as is

src/bin/list_store.rs
sdroege: No, if there is a different type than a bool stored in that column then things are completely broken. We should not silently ignore that

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+
+        let values: [&dyn ToValue; 8] = [
+            &d.fixed.to_value(),
+            &d.number.to_value(),
+            &d.severity.to_value(),
+            &d.description.to_value(),
+            &0u32.to_value(),
+            &icon_name.to_value(),
+            &false.to_value(),
+            &sensitive.to_value(),
```
sdroege: You don't have to call `to_value()` here, just taking the reference should be enough

sdroege: And maybe this would be nicer as a method on the `Data` struct?

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+    let mut fixed = model.get_value(&iter, COLUMN_FIXED).get::<bool>().unwrap();
+    fixed = !fixed;
+    model.set_value(&iter, COLUMN_FIXED as u32, &fixed.to_value());
+}
+
+fn add_columns(model: Rc<gtk::ListStore>, treeview: &TreeView) {
+    // Column for fixed toggles
+    {
+        let renderer = CellRendererToggle::new();
+        renderer.connect_toggled(move |w, path| fixed_toggled(&*model, w, path));
```
sdroege: This is so much nicer with `ListBox` and simply being able to bind a property of the model with the property on the widget :) Oh well!

src/bin/list_store.rs
sdroege: Not really, it can be even better: `let pulse = pulse.wrapped_add(1);` does exactly what is needed here :)

src/bin/list_store.rs
sdroege: Ugh, all the `ListStore`/etc API should really take a `&dyn ToValue` here instead of actual `&Value`. Then the `to_value()` call would be not needed.

src/bin/list_store.rs
```diff
@@ -0,0 +1,299 @@
...
+    let mut pulse = model.get_value(&iter, COLUMN_PULSE).get::<u32>().unwrap();
+    if pulse == u32::max_value() {
+        pulse = 0;
+    } else {
+        pulse += 1;
+    }
+
+    model.set_value(&iter, COLUMN_PULSE as u32, &pulse.to_value());
+    model.set_value(&iter, COLUMN_ACTIVE as u32, &true.to_value());
+
```
sdroege: You could do `model.set(&iter, &[COLUMN_PULSE, COUMN_ACTIVE], &[&pulse, &true]);` here

src/bin/list_store.rs
osa1: I actually copied from another example, but will upate.

src/bin/list_store.rs
osa1: Can you elaborate? This is porting a C example without much changes, but I'm interested in learning other ways of doing this.

src/bin/list_store.rs
sdroege: Best to also fix that example then :) See [here](https://github.com/sdroege/rustfest-rome18-gtk-gst-workshop/blob/a4a6cfd6fecd5485498a307991aefd6da8ca47cc/src/app.rs#L211) for a little bit cleaner pattern with all this, but not really needed for such a small example.

src/bin/list_store.rs
sdroege: That was just a comment about the uglyness of the `TreeView` / `ListStore` API, nothing wrong with your code.

Take a look at the `listbox_model.rs` example for something similar with `ListBox`. There the connection between the model and the UI is a simple property binding.

EPashkin:
@osa1 Thanks.
Now we have 4th list store example 😉

sdroege:
@osa1 please write a comment in the PRs once you addressed all comments (and of course if you have questions!), it's not clear otherwise if it's ready for another review or not :)

GuillaumeGomez:
Still working on this?

osa1:
> Now we have 4th list store example

What can I say.. lists are important ;-)

> Still working on this?

Will address comments later today.

osa1:
This is ready to go.

src/bin/list_store.rs
GuillaumeGomez: If all people were starting to add their names in the header, it wouldn't be sustainable much time. Again, there are git logs. The headers in those examples are only supposed to describe shortly what it is about.

src/bin/list_store.rs
sdroege: This one is still outstanding

src/bin/list_store.rs
```diff
@@ -0,0 +1,311 @@
...
+    treeview.set_search_column(Columns::Description as i32);
+
+    sw.add(&treeview);
+
+    add_columns(model.clone(), &treeview);
+
+    window.show_all();
+
+    let model = model.clone();
+    timeout_add(80, move || spinner_timeout(&model));
```
sdroege: Hm how does this compile? You don't import that function directly anymore via `gtk::*`, so it probably would have to be prefixed with `gtk::` now?

src/bin/list_store.rs
```diff
@@ -0,0 +1,311 @@
...
+    _w: &W,
+    path: gtk::TreePath,
+) {
+    let iter = model.get_iter(&path).unwrap();
+    let mut fixed = model
+        .get_value(&iter, Columns::Fixed as i32)
+        .get::<bool>()
+        .unwrap();
+    fixed = !fixed;
+    model.set_value(&iter, Columns::Fixed as u32, &fixed.to_value());
```
sdroege: Such a weird API: needs `u32` for the setter but `i32` for the getter :) Not your fault of course but...

See also https://blog.gtk.org/2019/02/08/report-from-the-gtk-hackfest-in-brussels/ the section about `GtkTreeView` btw. This is all hopefully going away completely with GTK4 and becomes a nicer API.

sdroege:
Looks good to me apart from
* https://github.com/gtk-rs/examples/pull/221/files#r255327253
* https://github.com/gtk-rs/examples/pull/221/files#r250937432
* https://github.com/gtk-rs/examples/pull/221#discussion_r255316835

Thanks :)

osa1:
Ready to merge.

GuillaumeGomez:
Thanks a lot!

PR #225, opened 2019-02-06
Update minimum versions and add 3.24 to the travis CI

sdroege:
@EPashkin Can you upload a 3.24 version of the GTK bundle for travis? Until that is there this will fail :)

@GuillaumeGomez Is there any reason why we don't use docker images of reasonable well up to date Linux distros for the CI?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,9 @@ addons:
     packages:
     - libgtk-3-dev
 env:
-  - GTK=3.4
+  - GTK=3.14
   - GTK=3.18
   - GTK=3.22.30
+  - GTK=3.24
 script:
   - ./build_travis.sh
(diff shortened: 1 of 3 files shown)


sdroege:
Also Travis seems broken, there's no travis build here :)

In any case, all the other PRs can be merged now @GuillaumeGomez. Then this one would become mostly green (apart from the GTK 3.24 build, see above).

GuillaumeGomez:
👍

EPashkin:
@sdroege Added branch https://github.com/EPashkin/gtk-bootstrap/tree/gtk3_24 with new version,
when https://travis-ci.org/EPashkin/gtk-bootstrap/builds/489669170 finished I add release.

EPashkin:
Versions from https://packages.ubuntu.com/disco/libgtk-3-0

EPashkin:
Now we have 3.14 for testing, but it build on gtk+-3.15.12.tar.xz and glib-2.43.4.tar.xz
due https://forums.freebsd.org/threads/gtk-3-update.52576/
so I prefer use Xenial Ubuntu if it works.

PR #226, opened 2019-02-06
Remove gtk::StatusIcon example

sdroege:
This API is deprecated (and thus removed from the bindings now)
in favour of gio::Notification.


GuillaumeGomez:
Still makes me sad. :'(

Let's wait for CI for this one!

sdroege:
Yeah, now CI should be green apart from the GTK 3.24 build :)

GuillaumeGomez:
The error for the 3.14 seems weird...

sdroege:
It means that there's only GLib 2.40 available there but we now need 2.42. The others succeed because they come with a newer version of GLib together with GTK.

GuillaumeGomez:
Euh... Ok? What do we do then? We can't really keep this situation...

EPashkin:
Restarted GTK=3.24, nightly passed

EPashkin:
Started build bootstrap for gtk3.14, it will be in https://github.com/EPashkin/gtk-bootstrap/releases if CI passed

sdroege:
Yeah we'll basically need to update the GTK binaries. Which will need another update for the build_travis.sh to point to the new version.

@EPashkin should I do that as part of this PR or do you want to send one?

sdroege:
> Started build bootstrap for gtk3.14, it will be in https://github.com/EPashkin/gtk-bootstrap/releases if CI passed

Failed for some reason that is not clear from the log: https://travis-ci.org/EPashkin/gtk-bootstrap/builds/489888040

sdroege:
@EPashkin We have similar failures in `glib` with the 3.22 build: https://travis-ci.org/gtk-rs/glib/jobs/489916043 . I don't understand where that gets its versions from.

GuillaumeGomez:
It's funny that it works for 3.24 but not for 3.14.

GuillaumeGomez:
That should we do in here? 3.14 is supposed to be supported by gtk-rs so testing it is a requirement.

sdroege:
> That should we do in here? 3.14 is supposed to be supported by gtk-rs so testing it is a requirement.

I don't know how the CI setup works exactly but it seems you get GTK binaries from @EPashkin's repo somehow? That should contain GLib 2.42 then, which is what I think he did for https://github.com/gtk-rs/examples/pull/226#issuecomment-461292240

GuillaumeGomez:
Can you fix it @EPashkin please? :)

sdroege:
unrelated, why don't we use docker images of reasonably recent Linux distros for the CI? So that we could simply use the GTK versions that are coming with those distro versions.

GuillaumeGomez:
If you have an idea to fix it, please go ahead. :)

sdroege:
I don't know how any of this CI stuff works, sorry :) but that's what I saw other projects doing

GuillaumeGomez:
I never used docker either. I can't really merge until this solution is solved, otherwise it'd mean that we'd have failing CI **forever**.

sdroege:
@EPashkin was working on it and it just waits for his GTK binaries for 3.14 to be finished.

EPashkin:
Seems building gtk3.14 not helps as we have problem on any linking attempt ever before building examples
we need lower version back to travis's 😢

GuillaumeGomez:
Urg<sup>2</sup>.

EPashkin:
Seems we need go down to gtk3.10
https://packages.ubuntu.com/trusty/libgtk-3-0

EPashkin:
@sdroege Can you try use Ubuntu Xenial in travis?
https://docs.travis-ci.com/user/reference/xenial/
https://docs.travis-ci.com/user/reference/overview/

sdroege:
Why don't we use the GTK binaries that you're building for the CI?

sdroege:
But that's also why I mentioned docker. Travis' Linux distro is simply shit and completely outdated, for proper testing you want to have a more meaningful environment set up :)

EPashkin:
Travis has gtk3.16 on Xenial, IMHO better use it that my gtk3.14 that build from different versions.
I don't have experience with docker and don't know any dockered CI.

sdroege:
> Travis has gtk3.16 on Xenial, IMHO better use it that my gtk3.14 that build from different versions.

But we also want to test the e.g. 3.24 configuration or not?

>  I don't have experience with docker and don't know any dockered CI.

I know how to use that for GitLab CI but that's about it.

sdroege:
I'll try the Xenial Travis later, but this does not seem like a long-term solution.

EPashkin:
@sdroege, @GuillaumeGomez 
Please don't start any new build for 30min, rebuilding 3.24 bootstrap

EPashkin:
3.24 bootstrap updated

sdroege:
Looks better

GuillaumeGomez:
Thanks a lot to both of you! I merge then. :)

PR #229, opened 2019-02-12
Build with the maximum versions required for each configuration on th…

sdroege:
…e CI of the dependant crates

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ addons:
   apt:
     packages:
     - libgtk-3-dev
+    - libmount-dev
 env:
   - GTK=3.14
   - GTK=3.18
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,12 +36,9 @@ git = "https://github.com/gtk-rs/gtk"
 
 [features]
 #default = ["gtk_3_22_30", "futures-stable", "subclassing"]
-gtk_3_16 = ["gtk/v3_16"]
-gtk_3_18 = ["gtk_3_16", "gtk/v3_18"] #for CI tools
-gtk_3_20 = ["gtk_3_18", "gtk/v3_20"] #for CI tools
-gtk_3_22 = ["gtk_3_20", "gtk/v3_22"] #for CI tools
-gtk_3_22_30 = ["gtk_3_22", "gtk/v3_22_30"] #for CI tools
-gtk_3_24 = ["gtk_3_22_30", "gtk/v3_24"] #for CI tools
+gtk_3_18 = ["gtk/v3_18", "gdk-pixbuf/v2_32", "gdk/v3_18", "gio/v2_46", "glib/v2_46", "pango/v1_38"] #for CI tools
+gtk_3_22_30 = ["gtk_3_18", "gtk/v3_22_30", "gdk-pixbuf/v2_36", "gdk/v3_22", "gio/v2_56", "glib/v2_56", "pango/v1_42"] #for CI tools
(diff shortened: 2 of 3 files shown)


GuillaumeGomez:
:+1:

sdroege:
@EPashkin why does the 3.22.30 bundle have GLib 2.55 and not 2.56, and the 3.24 bundle also instead of 2.58?

GuillaumeGomez:
Dark magic once again.

EPashkin:
@sdroege because on https://packages.ubuntu.com/bionic/libgtk-3-0 gtk3.22.30=>libglib2.0-0 (>= 2.55.2),
and for https://packages.ubuntu.com/disco/libgtk-3-0 too.
As I remember I has problem with glib 2.56

sdroege:
2.55 is an unstable development release though :) And you should ideally always use the releases that were part of the relvent GNOME release, which are the version numbers I mentioned above. The dependencies in Ubuntu are not necessarily correct.

EPashkin:
Ubuntu is fastest source that I can find, trying to build.
Also I want change packages name to just gtk-3.22.30

sdroege:
> Ubuntu is fastest source that I can find, trying to build.

Sure but that also has 2.56 here: https://packages.ubuntu.com/bionic/libglib2.0-0
My point is that those versions in the dependencies shouldn't be trusted, not that you shouldn't use Ubuntu :)

sdroege:
@EPashkin So what should we do here? Is there a problem with updating the versions as they are now in the `Cargo.toml`?

EPashkin:
I preparing bootstraps with version from cargo toml,
so IMHO better wait, as I test build here too.

EPashkin:
gtk-3.22.30 is done, 
on 3.24 I need time to understand how to include https://download.gnome.org/sources/atk/2.30/atk-2.30.0.tar.xz, it uses meson+ninja

sdroege:
> it uses meson+ninja

You can install meson via pip if your distro doesn't have it. That also brings ninja then.

Then do

```
mkdir builddir && cd builddir
meson --prefix=/foo/bar ..
ninja
ninja install
```

meson is like the configure part, so if you had more things set than prefix do it there.

EPashkin:
@sdroege Thanks, I missed that build dir need created before.

EPashkin:
appveyor still use `gtk_3_20`

sdroege:
> appveyor still use `gtk_3_20`

Oops. Will fix later :)

sdroege:
> Oops. Will fix later :)

Fixed now already. Should all become green once the bundles are ok

sdroege:
@EPashkin let me know when things should be retried or if you run into other problems :)

EPashkin:
@sdroege Thanks, but I can restart job yourself, while don't have write access to repo 😉

sdroege:
> @sdroege Thanks, but I can restart job yourself, while don't have write access to repo wink

Oh indeed. Interesting :)

EPashkin:
@sdroege Please add     `- libmount-dev` to travis.
If travis passed, please, not merge PR, package still not finished.

sdroege:
> @sdroege Please add `- libmount-dev` to travis.

Done

EPashkin:
3.24 passed.
Seems we need update all .travis again no add `libmount-dev` or install it in build.sh here.

sdroege:
>  Seems we need update all .travis again to add `libmount-dev` or install it in build.sh here.

`.travis.yml` everywhere. We can't install it in `build.sh`, and we don't have any template `.travis.yml` :( You send the PRs for that?

EPashkin:
Ok, I do it later today.

sdroege:
@GuillaumeGomez apart from that, this is ready to go and then we need new PRs for everything once again.

GuillaumeGomez:
> then we need new PRs for everything once again.

Ready, I am!

PR #232, opened 2019-02-21
Remove checking for gtk_3_10

EPashkin:
cc @GuillaumeGomez, @sdroege

### DIFF
--- a/src/bin/builder_basics.rs
+++ b/src/bin/builder_basics.rs
@@ -5,56 +5,40 @@
 extern crate gio;
 extern crate gtk;
 
-#[cfg(feature = "gtk_3_10")]
-mod example {
-    use gtk;
+use gio::prelude::*;
+use gtk::prelude::*;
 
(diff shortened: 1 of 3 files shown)


src/bin/builder_basics.rs
```diff
@@ -5,56 +5,40 @@
 extern crate gio;
 extern crate gtk;
 
-#[cfg(feature = "gtk_3_10")]
```
sdroege: Why did we do that inside the source code instead of just in Cargo.toml?! :)

sdroege:
:+1:

src/bin/builder_basics.rs
GuillaumeGomez: Young we were.

GuillaumeGomez:
Thanks!

EPashkin:
@sdroege Mostly because we don't add files to Cargo.toml and use auto-detection instead

PR #240, opened 2019-04-26
Port to futures 0.3

sdroege:
While this requires nightly, the API is now stabilized and will become part of Rust 1.36. IMHO we should merge this now so that it can all get wider testing.

This is currently marked as WIP because it uses https://github.com/gtk-rs/glib/pull/476 and https://github.com/gtk-rs/gir/pull/753 via Cargo patching

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ autobins = false
 [dependencies]
 chrono = "0.4"
 url = "1.4"
-futures-preview = { version = "0.2", optional = true }
+futures-preview = { version = "0.3.0-alpha", optional = true}
 
 [dependencies.atk]
 git = "https://github.com/gtk-rs/atk"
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -144,3 +145,6 @@ name = "treeview"
 
 [[bin]]
 name = "list_store"
+
+[patch.crates-io]
+futures-preview = { git = "https://github.com/rust-lang-nursery/futures-rs.git"}
```
sdroege: This is needed because of the `future::Either` type we use in the non-async/await example. Will send a new PR once that one is merged.

sdroege: I mean once that one is uploaded to crates.io :)

sdroege:
This should succeed now. Let's see :)

sdroege:
Yeah this one works

sdroege:
@GuillaumeGomez green, appveyour is still not fixed (unrelated to this PR) so will fail

GuillaumeGomez:
Should we merge it as is?

GuillaumeGomez:
To be more precise,

```toml
[patch.crates-io]
futures-preview = { git = "https://github.com/rust-lang-nursery/futures-rs.git"}
```

isn't making me very confident...

sdroege:
See comment above: https://github.com/gtk-rs/examples/pull/240#pullrequestreview-231125510

Alternatively I can also copy the implementation of `future::Either` into the example, but that does not seem very useful.

GuillaumeGomez:
True... Then let's get this merged as is. Thanks! :)

sdroege:
Thanks, triggering the glib PR and the gio master build then.

PR #241, opened 2019-05-02
Use futures from crates.io instead of git

sdroege:


sdroege:
@GuillaumeGomez all green

GuillaumeGomez:
That version number is completely wtf haha. Thanks!

PR #243, opened 2019-05-10
Add builders example

antoyo:


src/bin/builders.rs
```diff
@@ -0,0 +1,43 @@
...
+
+use std::env::args;
+
+fn build_ui(application: &gtk::Application) {
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title("First GTK+ Program");
+    window.set_border_width(10);
+    window.set_position(gtk::WindowPosition::Center);
+    window.set_default_size(350, 70);
```
sdroege: Would be nice to also have a builder for the above things :)

antoyo: There's no builder generated for `ApplicationWindow`.
To be able to do this, that would require the builder to generate methods for parent properties. I'll see if I can do that.

EPashkin:
@antoyo This PR better be against "pending" branch

sdroege:
@antoyo What's the status here? Would be good to have this updated against the pending branch and then merge it :)

src/bin/builders.rs
sdroege: ... which happened AFAIU now? :)

src/bin/builders.rs
```diff
@@ -0,0 +1,43 @@
...
+
+    window.add(&button);
+
+    window.show_all();
+}
+
+fn main() {
+    let application = gtk::Application::new("com.github.gtk-rs.examples.basic",
+                                            Default::default())
+                                       .expect("Initialization failed...");
```
sdroege: Also wants a run of `cargo fmt` :)

sdroege:
@GuillaumeGomez Good for me now :)

GuillaumeGomez:
Waiting for CIs in any case. :)

GuillaumeGomez:
Thanks @antoyo!

PR #245, opened 2019-05-12
Add example for dnd in a listbox with a listmodel

jsparber:
This is an example for implementing a queue or anything where the user
can reader a list via dnd. This includes also some animation for hidding
and showing rows


jsparber:
I still need to add more comments

sdroege:
Nice :) I'll review once the comments are there, let me know!

sdroege:
@jsparber What's the status of this? Do you plan to update it? :) Would be nice to get this merged as nothing really is missing here apart from some comments from what I understand?

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+    ($($n:ident),+ => move |$($p:tt),+| $body:expr) => (
+        {
+            $( let $n = $n.clone(); )+
+            move |$(clone!(@param $p),)+| $body
+        }
+    );
+}
+
+// upgrade weak reference or return
+#[macro_export]
```
sdroege: Why export?

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+    );
+
+    let window = gtk::ApplicationWindow::new(application);
+
+    window.set_title("ListBox Model Sample");
+    window.set_border_width(10);
+    window.set_position(gtk::WindowPosition::Center);
+    window.set_default_size(320, 480);
+
+    let window_weak = window.downgrade();
```
sdroege: Generally it's nicer to clean/downgrade things right before the closure where they're moved. Makes it easier to understand what they are and where they come from exactly.

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+                                               let model = upgrade_weak!(model_weak);
+                                               if let Some(data) = channel.take() {
+                                                   if let Some(source) = model.get_object(data.position) {
+                                                       if let Some(ref list) = widget.downcast_ref::<gtk::ListBox>() {
+                                                           if let Some(row) = list.get_row_at_y(y) {
+                                                           model.remove(data.position);
+                                                           let index = row.get_index() as u32;
+                                                           let alloc = row.get_allocation();
+
+                                                           let index = if y < alloc.y + alloc.height/2 {
```
sdroege: I believe you want to run `cargo fmt` once :)

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+                .expect("Failed to create row data")
+                .downcast()
+                .expect("Created row data is of wrong type")
+        }
+    }
+}
+
+// return the previous visible row
+fn get_previous_row(list: &gtk::ListBox, row: &gtk::ListBoxRow) -> Option<gtk::ListBoxRow> {
+    if let Some(w) = list.get_row_at_index(row.get_index() - 1) {
```
sdroege: What if `row.get_index() - 1 < 0`? No problem and it just returns `None`?

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+            Some(w)
+        }
+    } else {
+        None
+    }
+}
+
+// return the next visible row
+fn get_next_row(list: &gtk::ListBox, row: &gtk::ListBoxRow) -> Option<gtk::ListBoxRow> {
+    if let Some(w) = list.get_row_at_index(row.get_index() + 1) {
```
sdroege: What if `row.get_index() + 1 >= num_rows`?

src/bin/drag_and_drop_listbox_model.rs
```diff
@@ -0,0 +1,653 @@
...
+            let g = cairo::Context::new(&image);
+            //TODO use theme color as the background or add a temporarily style class
+            g.set_source_rgba(1.0, 1.0, 1.0, 0.8);
+            g.rectangle(0.0, 0.0, ww as f64, wh as f64);
+            g.fill();
+
+            w.draw(&g);
+
+            //TODO fix positioning
+            //https://stackoverflow.com/questions/24844489/how-to-use-gdk-device-get-position
```
sdroege: Might be worth to ask someone who knows GTK well to go through these FIXME/TODO comments :)

src/bin/drag_and_drop_listbox_model.rs
jsparber: Probably because i copied the macro.

src/bin/drag_and_drop_listbox_model.rs
jsparber: It returns null if there is no row on the specified index. 
https://gtk-rs.org/docs/gtk/trait.ListBoxExt.html#tymethod.get_row_at_index

src/bin/drag_and_drop_listbox_model.rs
jsparber: Yes it return None

src/bin/drag_and_drop_listbox_model.rs
jsparber: Those are there because i didn't had time to do it. But are actually not that complicated.

src/bin/drag_and_drop_listbox_model.rs
sdroege: Still seems like a good idea :)

jsparber: apparently rustfmt doesn't fix the spaces, actually it doesn't change the file add all

src/bin/drag_and_drop_listbox_model.rs
sdroege: Let's wait until that is done before merging then?

src/bin/drag_and_drop_listbox_model.rs
sdroege: That's curious. Don't worry about that then, I'll take a look once the code is done

PR #246, opened 2019-05-22
Update README

apiraino:
Just a little more detail that helped me (as a total newbie) to clear up things when trying to complile the examples.

make sense this stuff?

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,11 +6,10 @@ A few gtk-rs examples. To build, just do:
 > cargo build
 ```
 
-or to enable GTK 3.1x, 3.2x, all examples as well:
+or to enable GTK 3.x depending on the version needed by the example (check Cargo.toml `[features]` to see all specific GTK compile features available):
 
 ```Shell
 > cargo build --features gtk_3_18
-> cargo build --features gtk_3_22_30
 > cargo build --all-features
 ```
 
@@ -19,12 +18,14 @@ And then run the executables with:
 ``` Shell
 ./target/debug/EXAMPLE-NAME
 ```
+
 or with cargo run (repeating the compilation features used above), example:
+
 ``` Shell
 cargo run --all-features --bin EXAMPLE-NAME
 ```
 
-Please be sure to have installed all the required libraries before building examples (the list is available on [gtk](https://github.com/gtk-rs/gtk/).
+Please be sure to have installed all the required libraries before building examples (the list is available on the [gtk-rs](https://github.com/gtk-rs/gtk/) repository).
 
 ## LICENSE
 The gtk-rs examples repository is licensed under the MIT license, please refer to the LICENSE and COPYRIGHT files for more information.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,13 +6,25 @@ A few gtk-rs examples. To build, just do:
...
 ```
 
-or to enable GTK 3.10 examples as well:
+or to enable GTK 3.1x, 3.2x, all examples as well:
 
 ```Shell
-> cargo build --features gtk_3_10
+> cargo build --features gtk_3_18
+> cargo build --features gtk_3_22_30
+> cargo build --all-features
```
EPashkin: IMHO one version like 3.18 is enough.

GuillaumeGomez: I agree with @EPashkin in here.

README.md
```diff
@@ -6,13 +6,25 @@ A few gtk-rs examples. To build, just do:
 > cargo build
 ```
 
-or to enable GTK 3.10 examples as well:
+or to enable GTK 3.1x, 3.2x, all examples as well:
```
GuillaumeGomez: What about "or to enable GTK 3.x depending on the version needed by the example:"

README.md
```diff
@@ -6,13 +6,25 @@ A few gtk-rs examples. To build, just do:
...
+> cargo build --all-features
 ```
 
-And then run the executables. Please be sure to have installed all the required libraries before building examples (the list is available on [gtk](https://github.com/gtk-rs/gtk/).
+And then run the executables with:
+
+``` Shell
+./target/debug/EXAMPLE-NAME
+```
+or with cargo run (repeating the compilation features used above), example:
```
GuillaumeGomez: Please add an empty line before and after this one.

README.md
```diff
@@ -6,13 +6,25 @@ A few gtk-rs examples. To build, just do:
...
+``` Shell
+cargo run --all-features --bin EXAMPLE-NAME
+```
+
+Please be sure to have installed all the required libraries before building examples (the list is available on [gtk](https://github.com/gtk-rs/gtk/).
```
GuillaumeGomez: There is a missing closing paren at the end. :)

GuillaumeGomez: Also, I'd replace "\[gtk\]" with "\[gtk-rs\]" to make sure people don't confuse it with GTK official repository.

GuillaumeGomez:
Makes sense, indeed! Just a few nits and we're good to go.

README.md
apiraino: Ok, I'd like to reword that so the reader understands which are the options available and how to find it. E.g. "check Cargo.toml [features] to see all specific GTK compile features available"

GuillaumeGomez: To "enable them all", you can just enable the highest one. Also, in the cargo file, there is not only libraries versions handling but also features enablement. So I really like the first part of the sentence but way less the second one. ;)

apiraino:
done here, too (hoping to having correctly interpreted your suggestion)

GuillaumeGomez:
Thanks!

PR #248, opened 2019-05-23
Fix build after &Option<T> -> Option<&T> change

sdroege:
See https://github.com/gtk-rs/gir/pull/780

----

This will fail until https://github.com/gtk-rs/gtk/pull/816 is merged

### DIFF
--- a/src/bin/transparent_main_window.rs
+++ b/src/bin/transparent_main_window.rs
@@ -15,7 +15,7 @@ use std::env::args;
 
 fn build_ui(application: &gtk::Application) {
     let window = ApplicationWindow::new(application);
-    set_visual(&window, &None);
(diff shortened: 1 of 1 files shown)


src/bin/transparent_main_window.rs
```diff
@@ -15,7 +15,7 @@ use std::env::args;
 
 fn build_ui(application: &gtk::Application) {
     let window = ApplicationWindow::new(application);
-    set_visual(&window, &None);
+    set_visual(&window, None);
```
sdroege: This is btw one of the reasons why it's nicer to have the reference inside the `Option`

GuillaumeGomez:
Thanks!

PR #252, opened 2019-06-22
Use crates.io instead of git

GuillaumeGomez:
@EPashkin With this commit, next update shouldn't silently update dependencies without me noticing. :)


EPashkin:
Agree
👍

PR #255, opened 2019-07-24
Update minimum rust version

GuillaumeGomez:
We forgot to update it here as well. :smile: 

cc @EPashkin @sdroege


sdroege:
Also add another "stable" job while we're at it

PR #257, opened 2019-08-10
Update for new `glib::Value::get` signature

fengalin:
... in consequence of gtk-rs/glib#513

For the `Result<_, _>::expect` behaviour, please see the following discussion: https://github.com/gtk-rs/gir/pull/825#pullrequestreview-272474063. TLDR: the argument for `expect` should refer to the context (or the reason why it shouldn't panic), the error will be automatically added in the panic message (using the `Debug` trait unfortunately).

For the method names, please see: https://github.com/gtk-rs/glib/pull/513#pullrequestreview-271892055 and then https://github.com/gtk-rs/glib/pull/513#issuecomment-519547950. TLDR: `get` allows writing generic code, `get_some` is the best name we could think of to `get` the actual value for a non-optional `Type` without the need to `unwrap` an `Ok` output of `get`.

### DIFF
--- a/src/bin/list_store.rs
+++ b/src/bin/list_store.rs
@@ -7,6 +7,7 @@ use gtk::prelude::*;
 use std::env::args;
 use std::rc::Rc;
 
+#[derive(Debug)]
 #[repr(i32)]
 enum Columns {
     Fixed = 0,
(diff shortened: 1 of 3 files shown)


EPashkin:
Please, repase it to `pending` branch and `Edit` PR

EPashkin:
@fengalin Thanks

fengalin:
\o/

sdroege:
@GuillaumeGomez looks all good now :)

GuillaumeGomez:
Indeed! Thanks @fengalin !

PR #263, opened 2019-11-02
Minor fix in menu_bar_system

BO41:
triple slashes are used for function documentation.
`set_author` is called twice. I assume this is not needed

### DIFF
--- a/src/bin/menu_bar_system.rs
+++ b/src/bin/menu_bar_system.rs
@@ -58,7 +58,7 @@ fn build_system_menu(application: &gtk::Application) {
     application.set_menubar(Some(&menu_bar));
 }
 
-// This function creates "actions" which connect on the declared actions from the menu items.
+/// This function creates "actions" which connect on the declared actions from the menu items.
 fn add_actions(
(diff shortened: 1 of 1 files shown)


EPashkin:
@BO41 Thanks, 👍

src/bin/menu_bar_system.rs
```diff
@@ -58,7 +58,7 @@ fn build_system_menu(application: &gtk::Application) {
     application.set_menubar(Some(&menu_bar));
 }
 
-// This function creates "actions" which connect on the declared actions from the menu items.
+/// This function creates "actions" which connect on the declared actions from the menu items.
```
GuillaumeGomez: Why changing this to a doc comment?

BO41: well, it is a comment on a function. And making it a doc comment enables it be shown in IDEs and the docs.
Other people may be interested in the function of this function.
So why not use the already existing comment :)

GuillaumeGomez:
Thanks!

PR #264, opened 2019-11-02
Prelude import

GuillaumeGomez:
Depending on the discussion in https://github.com/gtk-rs/gio/pull/250, we might not want to merge this PR...

cc @sdroege @EPashkin

### DIFF
--- a/src/bin/tree_model_sort.rs
+++ b/src/bin/tree_model_sort.rs
@@ -6,8 +6,7 @@ extern crate gdk;
 extern crate gio;
 extern crate gtk;
 
-use gio::ApplicationExt;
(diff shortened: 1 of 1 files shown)


EPashkin:
👍

src/bin/tree_model_sort.rs
```diff
@@ -6,8 +6,7 @@ extern crate gdk;
 extern crate gio;
 extern crate gtk;
 
-use gio::ApplicationExt;
-use gio::ApplicationExtManual;
```
sdroege: This change is good nonetheless. We should not import traits manually but the prelude.

PR #265, opened 2019-11-04
Add Printing sample

MateusRodCosta:
For work I was doing a small GTK app that took some input and printed it. I was doing it in Rust with gtk-rs and, sadly, I didn't find much about how to do the printing part.
After a while looking at the GTK and gtk-rs docs and with the help of some IRC users I figured out how to do it.

After managing to do it, they suggested me to make a sample to add to this repo, so here it is.

This sample has a window with a header bar and three items: 2 Entry fields and a Button. Clicking the button starts up the print dialog with the contents of the Entry fields.

I also added the pangocairo crate, which is basically only used to render the pango Paragraph with `show_layout()`.

### DIFF
--- a/src/bin/printing.rs
+++ b/src/bin/printing.rs
@@ -15,6 +15,20 @@ use gtk::prelude::*;
 
 use std::env::args;
 
+// upgrade weak reference or return
+#[macro_export]
+macro_rules! upgrade_weak {
+    ($x:ident, $r:expr) => {{
+        match $x.upgrade() {
+            Some(o) => o,
+            None => return $r,
+        }
+    }};
+    ($x:ident) => {
+        upgrade_weak!($x, ())
+    };
+}
+
 fn print(window: &gtk::Window, value1: String, value2: String) {
     let print_operation = gtk::PrintOperation::new();
 
@@ -74,11 +88,12 @@ fn build_ui(application: &gtk::Application) {
         .get_object("buttonprint")
         .expect("Couldn't get buttonprint");
 
-    let window_clone = window.clone();
+    let weak_window = window.downgrade();
     button_print.connect_clicked(move |_| {
+        let window = upgrade_weak!(weak_window);
         let text1 = entry1.get_text().expect("Couldn't get text1").to_string();
         let text2 = entry2.get_text().expect("Couldn't get text2").to_string();
-        print(&window_clone, text1, text2);
+        print(&window, text1, text2);
     });
 
     window.show_all();
--- a/src/bin/printing.rs
+++ b/src/bin/printing.rs
@@ -1,0 +1,99 @@
+//! # Printing
+//!
+//! This sample reads text from two Entry fields,
+//! shows a print dialog and prints both texts one below
+//! the other.
+
+extern crate cairo;
+extern crate gio;
+extern crate gtk;
+extern crate pango;
+extern crate pangocairo;
+
+use gio::prelude::*;
+use gtk::prelude::*;
+
+use std::env::args;
+
+fn print(window: &gtk::Window, value1: String, value2: String) {
+    let print_operation = gtk::PrintOperation::new();
+
+    // Currently unused
+    // Could be used to check whether there was a success in printing
+    //let print_operation_result: gtk::PrintOperationResult;
+
+    print_operation.connect_begin_print(move |print_operation, _| {
+        // This sets the number of pages of the document.
+        // You most likely will calculate this, but for this example
+        // it's hardcoded as 1
+        print_operation.set_n_pages(1);
+    });
+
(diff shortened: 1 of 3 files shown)


src/bin/printing.rs
```diff
@@ -0,0 +1,99 @@
...
+use gtk::prelude::*;
+
+use std::env::args;
+
+fn print(window: &gtk::Window, value1: String, value2: String) {
+    let print_operation = gtk::PrintOperation::new();
+
+    // Currently unused
+    // Could be used to check whether there was a success in printing
+    //let print_operation_result: gtk::PrintOperationResult;
```
GuillaumeGomez: How would it be used?

src/bin/printing.rs
```diff
@@ -0,0 +1,99 @@
...
+        cairo.move_to(10.0, 10.0);
+        pangocairo::functions::show_layout(&cairo, &pango_layout);
+
+        //Draw text2 below text1
+        pango_layout.set_text(&value2);
+        cairo.rel_move_to(0.0, 20.0);
+        pangocairo::functions::show_layout(&cairo, &pango_layout);
+    });
+
+    //Open Print dialog setting up main window as its parent
```
GuillaumeGomez: Please add a whitespace after `//`.

src/bin/printing.rs
```diff
@@ -0,0 +1,99 @@
...
+            .create_pango_layout()
+            .expect("Couldn't create pango layout");
+        pango_layout.set_font_description(Option::from(&font_description));
+
+        // Draw text1
+        pango_layout.set_text(&value1);
+        cairo.move_to(10.0, 10.0);
+        pangocairo::functions::show_layout(&cairo, &pango_layout);
+
+        //Draw text2 below text1
```
GuillaumeGomez: Please add a whitespace after `//`.

src/bin/printing.rs
```diff
@@ -0,0 +1,99 @@
...
+    print_operation.connect_draw_page(move |_, print_context, _| {
+        let cairo = print_context
+            .get_cairo_context()
+            .expect("Couldn't get cairo context");
+
+        // This allows you to get the width of the page
+        // Currently unused in this example
+        //let width = print_context.get_width();
+
+        //Initi pango and set a font
```
GuillaumeGomez: Please add a whitespace after `//`.

Also typo on "Initi".

GuillaumeGomez:
Thanks! Just a few nits and it's good to go for me.

src/bin/printing.rs
MateusRodCosta: It seems from the docs that it's the return of the `PrintOperationExt` `run()` method that starts the printing, and it returns one of these:
```
Error
Apply
Cancel
InProgress
```

MateusRodCosta: I 'm now printing it with the println! macro.

src/bin/printing.rs
MateusRodCosta: Done

src/bin/printing.rs
MateusRodCosta: Done

src/bin/printing.rs
MateusRodCosta: Done

GuillaumeGomez:
Okay, looks good to me.

@sdroege @EPashkin: generally we tend to only accept new examples on the pending branch, should I change the target?

EPashkin:
@MateusRodCosta Thanks.

@GuillaumeGomez Yes, please edit PR. Although changes in  Cargo.toml (include pangocairo as git) will be needed

src/bin/printing.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    let window: gtk::Window = builder.get_object("window").expect("Couldn't get window");
+    window.set_application(Some(application));
+    let entry1: gtk::Entry = builder.get_object("entry1").expect("Couldn't get entry1");
+    let entry2: gtk::Entry = builder.get_object("entry2").expect("Couldn't get entry2");
+    let button_print: gtk::Button = builder
+        .get_object("buttonprint")
+        .expect("Couldn't get buttonprint");
+
+    let window_clone = window.clone();
```
sdroege: This causes a reference cycle (window -> button -> clicked closure -> window): you need to use weak references here.

sdroege:
> @sdroege @EPashkin: generally we tend to only accept new examples on the pending branch, should I change the target?

Looks good to me apart from the reference cycle, let's get it into `pending` once that's fixed.

src/bin/printing.rs
sdroege: @GuillaumeGomez This needs fixing still

sdroege:
:+1:

PR #266, opened 2019-11-11
Update to stable futures

sdroege:


GuillaumeGomez:
Thanks! Just waiting for the CI result to see if there are anything that might have been out of our wiseness.

PR #268, opened 2019-11-11
Update to the new borrowing-based futures read/write API in gio

sdroege:
Depends on https://github.com/gtk-rs/gio/pull/255


EPashkin:
much simpler 👍

sdroege:
See dependency PR. This does not work like this :)

PR #270, opened 2019-11-14
Clone macro

GuillaumeGomez:
So we currently have a few issues with the `clone` macro: since it only allows idents, it cannot take input like `x.y` or even `*x`. Updating to `expr` would be enough I guess but not sure if that will be this simple. :3

cc @EPashkin @sdroege @antoyo

### DIFF
--- a/src/bin/cairo_threads.rs
+++ b/src/bin/cairo_threads.rs
@@ -12,29 +12,12 @@ use std::time::Duration;
 
 use cairo::{Context, Format, ImageSurface};
 use gio::prelude::*;
+use glib::clone;
 use gtk::prelude::*;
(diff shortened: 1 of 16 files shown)


sdroege:
> Updating to `expr` would be enough I guess but not sure if that will be this simple. :3

Well it's not. `let x.y = foo;` is not valid syntax :) It's the same thing as with `self`: we require that the caller uses a valid variable name.

We could probably add a `compile_error!` with a better error message for that.

EPashkin:
@GuillaumeGomez Thanks.

GuillaumeGomez:
And finally working! \o/

PR #280, opened 2019-12-30
IconView Example

artonio:


artonio:
not sure why the build is failing, I'm sure I don't understand something, help please

GuillaumeGomez:
Thanks for the PR! However a few things need to be updated. :)

First, could you target the `pending` branch please? (it'll very certainly require to rebase or just copy/paste your code once you switched branch if you don't want to mess with `git`)

Second, please add a bit more comments. The examples are meant to be used for learning so explanations are always a big plus.

Third (and last): please run `cargo fmt` on your code. ;)

PR #288, opened 2020-02-06
Don't use recursive main loops for the dialogs

sdroege:
Recursive main loops can cause interesting reentrancy issues with e.g.
mutexes or refcells and are generally not a good idea.

See also https://github.com/gtk-rs/gtk/issues/952.

----

CC @sprang

### DIFF
--- a/src/bin/builder_basics.rs
+++ b/src/bin/builder_basics.rs
@@ -3,9 +3,11 @@
 //! This sample demonstrates how to use the builder with an imported glade file
 
 extern crate gio;
+extern crate glib;
 extern crate gtk;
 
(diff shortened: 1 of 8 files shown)


GuillaumeGomez:
This is a super weird bug...

sdroege:
> This is a super weird bug...

Well, this PR is a good idea regardless of the bug :)

GuillaumeGomez:
Indeed. I'll merge it once CI passed.

sdroege:
@GuillaumeGomez green now :)

GuillaumeGomez:
Merged now. :)

PR #290, opened 2020-02-09
Add example to show usage of a futures channel on the event loop.

windelbouwman:


src/bin/communication_thread.rs
```diff
@@ -0,0 +1,66 @@
+/*
+Example on how to use a communication thread alongside with the GUI thread.
```
EPashkin: Please use `//!`

windelbouwman: Okay, good idea.

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
+//! Example on how to use a communication thread alongside with the GUI thread.
+//! 
+//! Tricks used here:
+//! - Use a queue to show data on the GUI.
+//! - Run an async function on the GUI event loop.
+//! - Use a seperate thread to handle incoming data and put it into a queue.
```
sdroege:
```suggestion
//! - Use a seperate thread to handle incoming data and put it into a channel.
```

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
+//! Example on how to use a communication thread alongside with the GUI thread.
+//! 
+//! Tricks used here:
+//! - Use a queue to show data on the GUI.
```
sdroege:
```suggestion
//! - Use a channel to show data on the GUI.
```

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
...
+    application.run(&args().collect::<Vec<_>>());
+}
+
+fn build_ui(application: &gtk::Application) {
+    let window = ApplicationWindow::new(application);
+    let label = Label::new(None);
+    window.add(&label);
+
+    // Create a channel between communication thread and main event loop:
+    let (sender, mut receiver) = futures::channel::mpsc::channel::<String>(1000);
```
sdroege:
```suggestion
    let (sender, mut receiver) = futures::channel::mpsc::channel(1000);
```

The type should be possible to infer or not? :)

sdroege:
```suggestion
    let (sender, mut receiver) = mpsc::channel::<String>(1000);
```

Maybe in combination with `use futures::channel;` at the top?

src/bin/communication_thread.rs
EPashkin: seperate ?

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    // Spawn queue receive task on the main event loop:
+    let main_context = glib::MainContext::default();
+    let future = async move {
+        while let Some(item) = receiver.next().await {
```
sdroege: Like this using an async block over the main context channel has not that much advantage. They only really begin to shine if you run multiple async operations in them, i.e. have multiple awaits. Not sure what else to do here though :)

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
...
+        }
+    };
+    main_context.spawn_local(future);
+
+    start_communication_thread(sender);
+
+    window.show_all();
+}
+
+fn start_communication_thread(mut sender: futures::channel::mpsc::Sender<String>) {
```
sdroege:
```suggestion
fn start_communication_thread(mut sender: mpsc::Sender<String>) {
```

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
...
+    // Spawn separate thread to handle communication:
+    thread::spawn(move || {
+        let mut counter = 0;
+        loop {
+            // Instead of a counter, your application code will
+            // block here on TCP or serial communications.
```
sdroege: Ideally you'd do such things asynchronously though, unless your APIs or use-case does not allow for that :) Maybe a comment about that could be useful here that threads are not always the best choice for such things.

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,65 @@
...
+            let data = format!("Counter = {}!", counter);
+            println!("Thread received data: {}", data);
+            sender.try_send(data).unwrap();
```
sdroege: That can only happen if the channel is full or the receiver is gone. In case of the latter you probably want to shut down your thread instead of panicking, in case of the former print just print that data was consumed to slow and skip the iteration

src/bin/communication_thread.rs
windelbouwman: It infers `str` which does not have a size at compile time.

src/bin/communication_thread.rs
windelbouwman: Agreed, but I do not know what to do further and keep the example small at the same time. I'm hoping people familiar with async/await will know the advantage.

src/bin/communication_thread.rs
windelbouwman: Agreed, I'll make a comment about it.

src/bin/communication_thread.rs
sdroege: Curious, I would've thought that `start_communication_thread()` would've made the compiler know that it must be a `String`

src/bin/communication_thread.rs
sdroege: Ok :)

src/bin/communication_thread.rs
windelbouwman: Agreed, I added `is_full` and `is_disconnected` checks.

src/bin/communication_thread.rs
windelbouwman: Me too, but I assume the type is inferred from the first usage, which is in the `set_text` function, so then it is determined at `str` instead of `String`.

sdroege: Oh well :)

Cargo.toml
```diff
@@ -92,6 +92,10 @@ name = "clock"
 [[bin]]
 name = "clone_macro"
 
+[[bin]]
+name = "communication_thread"
+edition = "2018"
```
sdroege: Unrelated to this PR but we should maybe move all the examples to 2018 at this point :) @GuillaumeGomez @EPashkin opinions?

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,80 @@
...
+//!
+//! Tricks used here:
+//! - Use a channel to show data on the GUI.
+//! - Run an async function on the GUI event loop.
+//! - Use a separate thread to handle incoming data and put it into a channel.
+
+extern crate futures;
+extern crate gio;
+extern crate glib;
+extern crate gtk;
```
sdroege: You don't need any of the `extern crate` statements here with the 2018 edition

src/bin/communication_thread.rs
```diff
@@ -0,0 +1,80 @@
...
+    start_communication_thread(sender);
+
+    window.show_all();
+}
+
+fn start_communication_thread(mut sender: mpsc::Sender<String>) {
+    // Note that blocking I/O with threads can be prevented
+    // by using asynchronous code, which is often a better
+    // choice. For the sake of this example, we showcase the
+    // way to use a thread when there is not other option.
```
sdroege:
```suggestion
    // way to use a thread when there is no other option.
```

or

```suggestion
    // way to use a thread when there is not any other option.
```

GuillaumeGomez:
Thanks!

PR #294, opened 2020-02-16
Fix clone_macro example

### LINKED ISSUE #293: Clone_macros doesn't work
I've tried to use this macro (even the example) but it doesn't work. There is no error during compile time, but it does nothing on button click (the example doesn't even open). Does anyone have any idea?

GuillaumeGomez:
I'll then backport this on the pending branch. Fixes #293.

Full explanations: the `state2` variable needs to be cloned otherwise it won't exist anymore, which will prevent the closure to be actually "called" (it'll return early).

cc @sdroege @EPashkin

### DIFF
--- a/src/bin/clone_macro.rs
+++ b/src/bin/clone_macro.rs
@@ -34,7 +34,7 @@ fn main() {
     {
         let state2 = Rc::new(RefCell::new(State::new()));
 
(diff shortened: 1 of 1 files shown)


EPashkin:
👍

sdroege:
:+1:

PR #296, opened 2020-02-20
Replace destroy calls with close

GuillaumeGomez:
Part of https://github.com/gtk-rs/gtk/pull/958

cc @EPashkin @sdroege


sdroege:
:+1:

PR #297, opened 2020-03-01
Add example for GLArea

jmaargh:
This example uses a custom glium backend to allow glium to render to a GLArea widget. A somewhat-reusable `glium_backend` module is provided to handle the messier parts of this process.

This is based on #44, but somewhat refactored for more reusability, and developed/tested with up-to-date dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,9 @@ autobins = false
 chrono = "0.4"
 futures = "0.3"
 once_cell = "1.2.0"
+glium = { version = "0.26.0", optional = true }
+shared_library = { version = "0.1.9", optional = true }
+epoxy = { version = "0.1.0", optional = true }
 
 [dependencies.atk]
 git = "https://github.com/gtk-rs/atk"
@@ -52,6 +55,7 @@ optional = true
 gtk_3_18 = ["gtk/v3_18", "gdk-pixbuf/v2_32", "gdk/v3_18", "gio/v2_46", "glib/v2_46", "pango/v1_38"] #for CI tools
 gtk_3_22_30 = ["gtk_3_18", "gtk/v3_22_30", "gdk-pixbuf/v2_36", "gdk/v3_22", "gio/v2_56", "glib/v2_56", "pango/v1_42"] #for CI tools
 gtk_3_24 = ["gtk_3_22_30", "gtk/v3_24", "atk/v2_30", "gdk-pixbuf/v2_36_8", "gdk/v3_24", "gio/v2_58", "glib/v2_58"] #for CI tools
+glarea = ["gtk/v3_16", "glium", "shared_library", "epoxy"]
 
 [[bin]]
 name = "accessibility"
@@ -181,3 +185,7 @@ name = "printing"
 name = "gio_async_tls"
 required-features = ["async-tls"]
 edition = "2018"
+
+[[bin]]
+name = "glarea"
+required-features = ["glarea"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ version = "0.6"
 optional = true
 
 [features]
-#default = ["gtk_3_22_30"]
+#default = ["gtk_3_22_30", "glarea"]
 gtk_3_18 = ["gtk/v3_18", "gdk-pixbuf/v2_32", "gdk/v3_18", "gio/v2_46", "glib/v2_46", "pango/v1_38"] #for CI tools
 gtk_3_22_30 = ["gtk_3_18", "gtk/v3_22_30", "gdk-pixbuf/v2_36", "gdk/v3_22", "gio/v2_56", "glib/v2_56", "pango/v1_42"] #for CI tools
 gtk_3_24 = ["gtk_3_22_30", "gtk/v3_24", "atk/v2_30", "gdk-pixbuf/v2_36_8", "gdk/v3_24", "gio/v2_58", "glib/v2_58"] #for CI tools
--- a/src/bin/glarea/glium_backend.rs
+++ b/src/bin/glarea/glium_backend.rs
@@ -26,7 +26,7 @@ pub trait GliumRenderer {
     /// `render` signal callback for the GLArea.
     fn draw(
         &mut self,
-        frame: glium::Frame,
+        frame: &mut glium::Frame,
         gl_area: &gtk::GLArea,
         gl_context: &gdk::GLContext,
     ) -> gtk::Inhibit;
@@ -104,11 +104,22 @@ where
     pub fn render(&mut self, gl_area: &gtk::GLArea, gl_context: &gdk::GLContext) -> gtk::Inhibit {
         match &self.facade {
             Some(facade) => {
-                let frame = glium::Frame::new(
+                let mut frame = glium::Frame::new(
                     facade.context.clone(),
                     facade.context.get_framebuffer_dimensions(),
                 );
-                self.renderer.draw(frame, gl_area, gl_context)
+
+                let inhibit = self.renderer.draw(&mut frame, gl_area, gl_context);
+
+                match frame.finish() {
+                    Err(err) => {
+                        eprintln!("Error finishin frame: {:?}", err);
+                        return Default::default();
+                    }
+                    _ => (),
+                }
+
+                inhibit
             }
             None => Default::default(),
         }
--- a/src/bin/glarea/glium_backend.rs
+++ b/src/bin/glarea/glium_backend.rs
@@ -81,7 +81,14 @@ where
                 self.check_current_context,
                 Default::default(),
             )
-            .unwrap() // TODO: Better error handling`
+        };
+
+        let context = match context {
+            Ok(ctx) => ctx,
+            Err(err) => {
+                eprintln!("Error: unable to create glium Context: {}", err);
+                return;
+            }
         };
 
         let facade = Facade { context };
--- a/src/bin/glarea/glium_backend.rs
+++ b/src/bin/glarea/glium_backend.rs
@@ -1,6 +1,6 @@
 //! # glium_backend utility module
 //!
-//! This module provides some general utility typoes that allow glium to manage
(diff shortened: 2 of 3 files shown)


GuillaumeGomez:
Please target the `pending` branch. We'll review it once done. :)

jmaargh:
Re-targeted to `pending` :)

Cargo.toml
```diff
@@ -52,6 +55,7 @@ optional = true
 gtk_3_18 = ["gtk/v3_18", "gdk-pixbuf/v2_32", "gdk/v3_18", "gio/v2_46", "glib/v2_46", "pango/v1_38"] #for CI tools
 gtk_3_22_30 = ["gtk_3_18", "gtk/v3_22_30", "gdk-pixbuf/v2_36", "gdk/v3_22", "gio/v2_56", "glib/v2_56", "pango/v1_42"] #for CI tools
 gtk_3_24 = ["gtk_3_22_30", "gtk/v3_24", "atk/v2_30", "gdk-pixbuf/v2_36_8", "gdk/v3_24", "gio/v2_58", "glib/v2_58"] #for CI tools
+glarea = ["gtk/v3_16", "glium", "shared_library", "epoxy"]
```
EPashkin: Please add glarea to commented `default`

EPashkin:
On windows it builds fine, but not works with "function was not loaded".
I think that it not stopper, as glium tests fails for me too,
seems my system missed something but I don't have time to investigate.


@GuillaumeGomez Please do "squash" on merge

GuillaumeGomez:
Ok, cleaned up the commits. Reviewing now.

src/bin/glarea/glium_backend.rs
```diff
@@ -0,0 +1,170 @@
+//! # glium_backend utility module
+//!
+//! This module provides some general utility typoes that allow glium to manage
```
GuillaumeGomez: "This module provides some general utility typoes that allow glium to manage" -> "This module provides some general utility types which allow glium to manage"

src/bin/glarea/glium_backend.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// Utility function showing how to use a GlAreaGliumHook to connect a GLArea to
+/// glium via a GliumRenderer.
+pub fn hook_to_renderer<R>(renderer: R, gl_area: &gtk::GLArea, check_current_context: bool)
+where
+    R: GliumRenderer + 'static,
+{
+    let hook = GlAreaGliumHook::new(renderer, check_current_context);
+
+    gl_area.connect_realize(glib::clone!(@strong hook => move |gl_area| {
```
GuillaumeGomez: `@strong` -> `@weak`

src/bin/glarea/glium_backend.rs
```diff
@@ -0,0 +1,170 @@
...
+        hook.borrow_mut().realize(gl_area);
+    }));
+
+    gl_area.connect_unrealize(glib::clone!(@strong hook => move |gl_area| {
```
GuillaumeGomez: `@strong` -> `@weak`

src/bin/glarea/glium_backend.rs
```diff
@@ -0,0 +1,170 @@
...
+        hook.borrow_mut().unrealize(gl_area);
+    }));
+
+    gl_area.connect_render(glib::clone!(@strong hook => move |gl_area, gl_context| {
```
GuillaumeGomez: No need to use the `clone!` macro here.

src/bin/glarea/glium_backend.rs
```diff
@@ -0,0 +1,170 @@
...
+    }
+
+    pub fn realize(&mut self, gl_area: &gtk::GLArea) {
+        let context = unsafe {
+            glium::backend::Context::new::<_>(
+                Backend::new(gl_area.clone()),
+                self.check_current_context,
+                Default::default(),
+            )
+            .unwrap() // TODO: Better error handling`
```
GuillaumeGomez: TODO :)

src/bin/glarea/main.rs
```diff
@@ -0,0 +1,145 @@
...
+use gio::prelude::*;
+use glium::Surface;
+use gtk::prelude::*;
+use gtk::WidgetExt;
+
+mod glium_backend;
+
+#[derive(Copy, Clone)]
+struct Vertex {
+    position: [f32; 2],
```
GuillaumeGomez: It's nitpicking but why not creating a `Position` type and a `Color` type?

src/bin/glarea/main.rs
```diff
@@ -0,0 +1,145 @@
...
+        in vec3 vertex_color;
+
+        out vec4 color;
+
+        void main() {
+            color = vec4(vertex_color, 1.0);
+        }"#;
+
+        let program =
+            glium::Program::from_source(facade, vert_shader_src, frag_shader_src, None).unwrap();
```
GuillaumeGomez: `unwrap`

src/bin/glarea/main.rs
```diff
@@ -0,0 +1,145 @@
...
+            frame
+                .draw(
+                    &triangle.vertex_buffer,
+                    &triangle.indices,
+                    &triangle.program,
+                    &glium::uniforms::EmptyUniforms,
+                    &Default::default(),
+                )
+                .unwrap();
+            frame.finish().unwrap();
```
GuillaumeGomez: `unwrap`

Cargo.toml
jmaargh: done

src/bin/glarea/glium_backend.rs
jmaargh: Appropriate typo, fixed

src/bin/glarea/glium_backend.rs
jmaargh: Done

src/bin/glarea/glium_backend.rs
jmaargh: Done

src/bin/glarea/glium_backend.rs
jmaargh: Done

src/bin/glarea/glium_backend.rs
jmaargh: Replaced just with stderr logging, since the only more robust error handling mechanisms I could think of would distract from the example.

src/bin/glarea/main.rs
jmaargh: Not done, because I think it's overkill. Also, I'd have to check if it played nicely with the `implement_vertex!` macro. Also, it's important that the layout of these types match in the shaders, and it's more cumbersome to verify that if they're newtypes.

src/bin/glarea/main.rs
jmaargh: done

src/bin/glarea/main.rs
jmaargh: done

jmaargh:
Fixed all issues.

Error "handling" is lazy, because I can't think of anything more robust that doesn't feel overkill for a simple example.

GuillaumeGomez:
I fixed your commit history and you broke it all again :-/

jmaargh:
Sorry, I think we're seeing different histories (presumably because you can't change the branch that's in my fork?) When you say "fixed" do you mean rebased to `gtk-rs:pending`?

GuillaumeGomez:
No, I force-pushed on your branch. I'll let you fix it this time. :)

jmaargh:
Fixed the history

src/bin/glarea/main.rs
GuillaumeGomez: Ok!

GuillaumeGomez:
Code looks good to me. I just need to actually launch it to confirm and then it's all good for me.

sdroege:
What's the status here? @GuillaumeGomez ?

GuillaumeGomez:
When closing the window, I get:

```
glarea: ../src/dispatch_common.c:863: epoxy_get_proc_address: Assertion `0 && "Couldn't find current GLX or EGL context.\n"' failed.
Aborted (core dumped)
```

Could you fix this please? Once done I'll merge.

jmaargh:
I don't remember seeing this before, but now I see the same assertion failure on exit (I have done a clean OS install since then, so who knows what could have changed).

Backtrace shows it's from epoxy not finding the context when `Gl::DeleteVertexArrays` is called as a result of `self.triangle = None;`. I've played around a bit and can't figure out exactly why the context seems to have disappeared before this point. In particular, the GL context is held by an `Rc` in the `Facade` object, which should still be in scope at that point.

[This](https://github.com/GNOME/gtkmm/blob/master/demos/gtk-demo/example_glarea.cc#L86) C++ example seems to suggest that you need to call (the equivalent of) `connect_unrealize` with `after = false`, but I couldn't get the rust API to let me do that (not even manually calling the underlying `glib` methods, mostly because their use is undocumented).

Without help, I'm not sure this is an error I can track down in a reasonable amount of time.

sdroege:
> [This](https://github.com/GNOME/gtkmm/blob/master/demos/gtk-demo/example_glarea.cc#L86) C++ example seems to suggest that you need to call (the equivalent of) `connect_unrealize` with `after = false`, but I couldn't get the rust API to let me do that (not even manually calling the underlying `glib` methods, mostly because their use is undocumented).

Right now you could use `area.connect("unrealize", false, |args| { ... })` from `glib::ObjectExt`

jmaargh:
I successfully used `connect_local` from `glib::ObjectExt` but just got the same result, with the same stack trace.

Somewhere after the window close request is triggered, something is causing `epoxy_current_context_is_glx` to fail to find a GLX or EGL context, when it was finding one fine before. As far as I can tell, the logic of this PR is sound -- unload GL resources on `GLArea` unrealize. The fact that something else within GTK/Glib/libepoxy/GTK-rs causing `libepoxy` to get confused before this can happen doesn't appear to be something this example can work around with anything other than hacks.

PR #308, opened 2020-04-25
basic_subclass: fix name of our GObject subclasss type

### LINKED ISSUE #307: in basic_subclass exmaple, the Private struct is exposed to gtk
As discussed in https://github.com/gtk-rs/examples/issues/279#issue-542074234

>  Shouldn't be it SimpleWindow instead of the SimpleWindowPrivate in the inspector?

https://user-images.githubusercontent.com/6001248/71407718-22aedd80-263c-11ea-9743-06d5f58770f1.png

alatiera:
Close #307

### DIFF
--- a/src/bin/basic_subclass.rs
+++ b/src/bin/basic_subclass.rs
@@ -42,7 +42,7 @@ pub struct SimpleWindowPrivate {
 }
 
 impl ObjectSubclass for SimpleWindowPrivate {
-    const NAME: &'static str = "SimpleWindowPrivate";
+    const NAME: &'static str = "SimpleWindow";
     type ParentType = gtk::ApplicationWindow;
     type Instance = subclass::simple::InstanceStruct<Self>;
     type Class = subclass::simple::ClassStruct<Self>;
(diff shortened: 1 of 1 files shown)


alatiera:
cc @sdroege I think this is correct but not entirerly sure. The name constant there corresponds to what we would set with ClassName in C right?

sdroege:
> The name constant there corresponds to what we would set with ClassName in C right?

Yes. That's why I usually do things different than you did here. An `imp` module with `struct Foo` for the implementation, and outside the `imp` module a `struct Foo` for the public type. This way you don't have the confusion about type names and can distinguish between the two via `imp::Foo` and `super::Foo`.

alatiera:
done as well, I think its from when I initially had implemented the functions on the public struc.

sdroege:
Looks good to me, @GuillaumeGomez please merge when green :)

GuillaumeGomez:
:+1:

sdroege:
@GuillaumeGomez green

GuillaumeGomez:
Thanks @alatiera !

PR #314, opened 2020-06-15
Update new_with_XXX/new_from_XXX function names

sdroege:


GuillaumeGomez:
Merging so other CIs will pass. Thanks!

PR #329, opened 2020-09-30
Remove deprecated functions and unused imports

Hofer-Julian:


Hofer-Julian:
Not sure why build fails here, but it doesn't seem to be related to my commit.

sdroege:
This was already done in the meantime

PR #331, opened 2020-10-04
Use https for all links

dns2utf8:
Hi all

This is a followup to https://github.com/gtk-rs/gtk/pull/1054

Cheers,
Stefan

### DIFF
--- a/src/bin/cairo_png.rs
+++ b/src/bin/cairo_png.rs
@@ -2,7 +2,7 @@
 //!
 //! This sample demonstrates how to create `ImageSurface`, draw on it
 //! and then save result to PNG file.
-//! Analog of C# example http://www.mgsloan.com/cairo_tut/stroke.cs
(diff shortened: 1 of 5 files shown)


src/bin/cairo_png.rs
```diff
@@ -2,7 +2,7 @@
 //!
 //! This sample demonstrates how to create `ImageSurface`, draw on it
 //! and then save result to PNG file.
-//! Analog of C# example http://www.mgsloan.com/cairo_tut/stroke.cs
+//! Analog of C# example https://web.archive.org/web/20130729155927/http://www.mgsloan.com/cairo_tut/stroke.cs
```
GuillaumeGomez, line 5: You failed your replace here. ;)

dns2utf8: Sadly the original site has been restructured  and the only reference I could find was in the Web Archive.
But the archive URL is with TLS :upside_down_face:

PR #335, opened 2020-10-17
Update listbox_model.rs for new subclass property API

sdroege:
See https://github.com/gtk-rs/glib/pull/685


sdroege:
Superseded by https://github.com/gtk-rs/gtk-rs/pull/31

